PR #3, opened 2018-08-28
add inject reference and SCSS styling

zparnold:
This PR adds:

- The command line flags for `linkerd inject` and explanation for those flags
- SCSS Styling for Tables, and code syntax highlighting

### DIFF
--- a/linkerd.io/static/scss/bootstrap/_tables.scss
+++ b/linkerd.io/static/scss/bootstrap/_tables.scss
@@ -2,7 +2,7 @@
 // Basic Bootstrap table
 //
 
-.table {
+table {
   width: 100%;
   margin-bottom: $spacer;
   background-color: $table-bg; // Reset for nesting within parents with `background-color`.
@@ -12,6 +12,7 @@
     padding: $table-cell-padding;
     vertical-align: top;
     border-top: $table-border-width solid $table-border-color;
+    min-width: $table-cell-min-width;
   }
 
   thead th {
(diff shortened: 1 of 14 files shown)


wmorgan:
@zparnold delivers! Thank you! Don't forget your DCO signoff :)

zparnold:
@wmorgan I have Signed the commit now. 😄 Thanks!

linkerd.io/static/scss/bootstrap/_tables.scss
```diff
@@ -2,7 +2,7 @@
 // Basic Bootstrap table
 //
 
-.table {
+table {
```
grampelberg: We're trying to treat all of the `bootstrap` stuff as vendored. You can add classes to markdown elements like so:

```markdown
+++
markup = "mmark"
+++

{.table .table-striped .table-bordered .table-responsive .table-min-width}
| Foo | Bar |
| ---  | ---  |
```

The min-width class can be moved to `index.scss` as a separate class that gets added to that table.

linkerd.io/static/scss/bootstrap/_variables.scss
```diff
@@ -935,7 +935,8 @@ $close-text-shadow:                 0 1px 0 $white !default;
 // Code
 
 $code-font-size:                    87.5% !default;
-$code-color:                        $pink !default;
+$code-color:                        $gray-800 !default;
```
grampelberg: I do like this over the current pink (who thought that was a good idea?). Mind moving the updates to `custom.scss` and `index.scss`? There's going to be a weird import order issue, so it might be worth moving all the `$gray-xx` SCSS code to `custom.scss` to get around that.

PR #15, opened 2018-09-17
Add stable and edge selectors

grampelberg:


linkerd.io/layouts/partials/sidebar.html
```diff
@@ -16,7 +16,11 @@ <h6 class="dropdown-header">Linkerd 1.x</h6>
...
       {{ .page.Site.Params.latest_release_version }}
     </a>
-    <h6 class="dropdown-header">Linkerd 2.x</h6>
+    <h6 class="dropdown-header">Linkerd 2.x Stable</h6>
+    <a class="dropdown-item" href="/2/overview">
+      2.0.0
+    </a>
+    <h6 class="dropdown-header">Linkerd 2.x Edge</h6>
     <a class="dropdown-item" href="/2/overview/">
       {{ .page.Site.Params.l5d2_release_version }}
```
dadjeibaah: It seems like this adds the string "L5D2_LATEST_VERSION" when I build it locally or is there some special env var magic that happens when we deploy the site?

grampelberg: Yup, https://github.com/linkerd/website/blob/master/Makefile#L75

dadjeibaah: Cool!

klingerf:
I think we should drop this PR and only publish docs for stable releases. Relates to #19.

PR #20, opened 2018-09-17
Update to use linkerd2 stats

grampelberg:


grampelberg:
@adleong how are you this much of a ninja?

PR #37, opened 2018-09-27
Add nodevoto manifest for install

grampelberg:


run.linkerd.io/public/nodevoto.yml
```diff
@@ -0,0 +1,179 @@
...
+      labels:
+        app: gif-svc
+    spec:
+      containers:
+      - env:
+        - name: GRPC_PORT
+          value: "8080"
+        - name: SVC_NAME
+          value: gif
+        image: sourishkrout/nodevoto:v4
```
klingerf: Random question: should we think about hosting this image in the buoyantio org on dockerhub? Not a blocker for this PR, just curious.

grampelberg: Yup, I've not dug into "productizing" this quite yet.

PR #51, opened 2018-10-24
Add booksapp config

klingerf:
This adds a Kubernetes config for installing the [books demo app](https://github.com/BuoyantIO/booksapp).


grampelberg:
It works, a fine work of art!

PR #61, opened 2018-11-21
Add instructions for ingress and why it doesn't work

grampelberg:


grampelberg:
@olix0r mind checking my explanation for why it doesn't work and make sure I'm correct?

Capitrium:
```
Note: it is not possible to rewrite the header in this way for the default
backend. Because of this, if you inject Linkerd into your Nginx ingress
controller's pod, the default backend will not be usable.
```

Just my 2c - this can also be handled, just not using ingress annotations:

The nginx-ingress-controller can be provided a [custom configuration template](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/custom-template.md) that can be used to rewrite the `Host` header for requests to the default backend.

In the latest version (which I have not yet tested myself), the `Host` header appears to get set to the value of the `$best_http_host` nginx template variable [here](https://github.com/kubernetes/ingress-nginx/blob/nginx-0.20.0/rootfs/etc/nginx/template/nginx.tmpl#L678). Replacing `$best_http_host` with `default-backend.ingress-nginx.svc.cluster.local` should make the default backend usable.

In 0.18.0 (which I have tested), the `Host` header is set [here](https://github.com/kubernetes/ingress-nginx/blob/nginx-0.18.0/rootfs/etc/nginx/template/nginx.tmpl#L1027).

grampelberg:
@Capitrium yeah, the only solution I could come up with was changing the template as well. That didn't feel like something that should go into a howto doc (have you seen that template, I thought *my* templates were complex).

KirbySSmith:
I was able to get the default backend working with nginx by setting controller.proxySetHeaders on the nginx helm chart. 
Adding the header to the ingress controller also means that the header does not have to be set on every ingress manifest. The helm chart does not have any way to set the gRPC header.
https://github.com/helm/charts/tree/258cd28472554d84fc206243074bbc11da3b2316/stable/nginx-ingress 
```
controller:
  proxySetHeaders:
    l5d-dst-override: $service_name.$namespace.svc.cluster.local:$service_port
```

grampelberg:
Woah, cool!

PR #64, opened 2018-11-22
"choose your platform" getting started

wmorgan:
@grampelberg I am interested in your thoughts on these changes.

- I took the choose-your-platform page that was in docs, and moved it to /choose-your-platform (with a little tweaking).
- I made the homepage "getting started" button back to one button, which points to that page ^ .
- I've made the /docs page just go to /2/overview for now, but I think in a later branch this could possibly be a page that talks about 1.x and 2.x and points users down the right path.

(I also made some more homepage wordsmithing since I was in there, but those can be done separately.)

I think this will all serve to clarify the state of 1.x vs 2.x platform support.


wmorgan:
As usual, could use some CSS help.

grampelberg:
@wmorgan looks like you missed the signoff ... somehow.

wmorgan:
@grampelberg would love one last pass from you

grampelberg:
@wmorgan merged, it looks good enough (tm) for me. I need to think about the overview changes a little bit more and make sure they fit into the whole picture.

linkerd.io/layouts/index.html
```diff
@@ -4,10 +4,9 @@
       <div class="row">
         <div class="col-md-7 pb-3">
           <h1>Ultralight service mesh for Kubernetes and beyond</h1>
-          <p>Linkerd gives you global observability, reliabilty, and security for your<br/> microservices &mdash; with no code changes required.</p>
+          <p>Linkerd gives you observability, reliabilty, and security for your<br/> microservices &mdash; with no code change required.</p>
```
klingerf: @wmorgan Sorry for chiming in late here. Was "with no code change required" intentional here? The previous "with no code changes required" sounds better to me.

wmorgan: Agreed, I'm fixing this in a followup branch

PR #65, opened 2018-11-27
Use simpler port-forwarding in getting started guide

klingerf:
This should hopefully help to simplify the process of viewing the emojivoto app in your web browser.


grampelberg:
@klingerf do you know what version of `kubectl` this landed in? I'd thought about going this route but got concerned how many folks would have the wrong version.

klingerf:
@grampelberg Ahh, good point. Looks like it was added in v1.10.0, which was released in March. I'm happy to discard this change if we think some folks are still on 1.9 or earlier.

grampelberg:
@klingerf what do you think about about waiting until 1.9 is EOL'd? Should be with 1.13 if I read the versioning docs correctly.

klingerf:
Yep, sounds good to me. Can leave this open for awhile if nobody minds.

PR #70, opened 2018-11-29
Modify Prometheus docs to use Federation

siggy:
The existing Prometheus docs instructed users to collect from Linkerd
proxies directly.

Instead provides instructions for using Prometheus federation, a simpler
and more robust approach.

screen shot 2018-11-29 at 2 03 28 pm

screen shot 2018-11-29 at 2 03 38 pm


siggy:
@wmorgan good catch on line length wrapping, fixed a couple.

PR #77, opened 2018-12-06
Update to latest booksapp config

klingerf:
Syncing this up with what's on master in the buoyantio/booksapp repo.


grampelberg:
@klingerf mind adding the swagger profiles as well?

grampelberg:
I'd been doing it on a per-app basis (`/emojivoto/spec.swagger`), but this works. We can move it around later if it becomes an issue.

klingerf:
I like the per-app suggestion. Have moved them to the `booksapp` subdir.

PR #81, opened 2018-12-07
Update booksapp config to use in-memory storage

klingerf:
Update the public booksapp.yml to install the app with sqlite3 backends instead of mysql. This drastically improves startup times, but means that we can only run 1 replica each of the books and authors deployments.


klingerf:
@grampelberg The mysql configs live here:

https://github.com/BuoyantIO/booksapp/tree/master/k8s

I wasn't planning on mirroring them on run.linkerd.io, but could definitely do that if you prefer.

grampelberg:
Nope, that's great. I got confused as to what repo I was looking at.

PR #134, opened 2019-02-06
minor tweaks to the Getting Started guide

wmorgan:
I went through this with a non-technical user and they stumbled
on one or two things, which I've tried to correct below.

### DIFF
--- a/linkerd.io/content/2/getting-started/_index.md
+++ b/linkerd.io/content/2/getting-started/_index.md
@@ -184,8 +184,8 @@ kubectl -n emojivoto port-forward \
   8080:80
 ```
 
-Now visit [http://localhost](http://localhost). Voila! The emojivoto app in its
-glory.
+Now visit [http://localhost](http://localhost:8080). Voila! The emojivoto app
+in all its glory.
 
 Clicking around, you might notice that some parts of the application are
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/getting-started/_index.md
```diff
@@ -172,38 +172,41 @@ curl -sL https://run.linkerd.io/emojivoto.yml \
...
 ```
 
-You might notice that some parts of the application are broken! If you were to
-inspect your handy local Kubernetes dashboard, you wouldn’t see very much of
-interest --- as far as Kubernetes is concerned, the app is running just fine.
-This is a very common situation! Kubernetes understands whether your pods are
-running, but not whether they are responding properly. Check out the
-[debugging example](../debugging-an-app) if you're interested in how to figure
-out exactly what is wrong.
+Now visit [http://localhost](http://localhost). Voila! The emojivoto app in its
```
grampelberg: This would have to be :8080 (and is confusing based off the docker desktop instructions). Maybe we should update emojivoto to listen on 8080? That would solve a PSP issue as well.

wmorgan: Good point. Updated doc. I'll leave the emojivoto thing for later ticketing

PR #110, opened 2019-01-25
Add a `project` label to all booksapp resources

olix0r:
This enables easy selection of all resources related to this demo.

Especially:

    kubectl delete all -l project=booksapp

PR #143, opened 2019-02-08
Fix missing label in booksapp config

klingerf:
This is a follow-up to #110, which adds a project=booksapp label to the traffic pod, since I noticed it was missing. Relates to BuoyantIO/booksapp#13.


klingerf:
Closing in favor of #144.

PR #145, opened 2019-02-09
move feature docs under a "Features" section

wmorgan:
Move TLS, CNI, HA, protocol detection, proxy injection, retries and timouts,
service profiles, and supported protocol content under a Features section.

Clean up protocol detection doc, add a stub HTTP, gRPC, and HTTP/2 doc

Future work: add a TOC to the Features index page, move that content into a separate dashboard, and various other cleanups.

PR #151, opened 2019-02-11
Fix ordering on feature index page

klingerf:
This is a follow-up to #145, which setup a new features index page. The index page wasn't really using the `sectiontoc` shortcode correctly, so items were showing up out of order, and the features index page itself was being included in the list of features. This should fix it. Adding a `description` tag on each of the feature pages would also make the index page a lot prettier. I can add that as part of this branch if nobody else is working on it.

Before this branch, on current master, the feature index page looks like this:

screen shot 2019-02-11 at 1 18 13 pm

And after this branch, it looks like this:

screen shot 2019-02-11 at 1 19 50 pm

### DIFF
--- a/linkerd.io/content/2/features/_index.md
+++ b/linkerd.io/content/2/features/_index.md
@@ -1,10 +1,10 @@
 +++
 date = "2018-09-10T12:00:00-07:00"
-title = "Features"
+title = "Overview"
 [menu.l5d2docs]
   name = "Features"
+  identifier = "features"
   weight = 3
 +++
 
 {{% sectiontoc "cli" %}}
-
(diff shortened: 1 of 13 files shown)


linkerd.io/content/2/features/_index.md
```diff
@@ -1,10 +1,10 @@
 +++
 date = "2018-09-10T12:00:00-07:00"
-title = "Features"
+title = "Overview"
```
wmorgan: I preferred Features but I'm not dead set on it.

klingerf: "Overview" is required in order for the `sectiontoc` shorthand to omit this page from the list of feature pages. We could fix our shorthand, but I'd rather not tackle that as part of this branch.

wmorgan: Ohhhhh, TIL

klingerf:
I took a first pass at description tags as part of this branch. We can improve these later in a follow-up branch, but for now it looks like this:

image

wmorgan:
Looks great!

PR #171, opened 2019-02-15
Add instructions for Traefik

grampelberg:


linkerd.io/resources/_gen/assets/scss/scss/index.scss_5ad6f408b0e3e473c748aac88af0ea18.content
```diff
@@ -0,0 +1,6508 @@
+/*!
```
klingerf: Hmm, should this file be in this diff? Seems like maybe it was checked in by accident?

grampelberg: *grumble* no. Netlify is using an older version of hugo that can't do this generation. So, for the time being we need to check these generated files in until they fix it on their end.

klingerf: Ahhh, ok, hmm. Does every update to the docs require a change to the generated files? Or rather, what is it about the docs change you're making in this branch that produces a new generated file? Is it a result of the shortcode that you're adding?

grampelberg: That's a good question, it doesn't show up in the file anywhere...

grampelberg: Looks like it does only change on css changes ... there's just a funky state right now that I need to figure out. Updated shrink the changes based off a rebase on master.

PR #172, opened 2019-02-18
Update to have pages for all CLI commands

grampelberg:


grampelberg:
@klingerf do you have any good ideas on how to integrate this into the release process? Right now you just need to remember to do it.

PR #210, opened 2019-03-04
emojivoto: Add service accounts

olix0r:
Though emojivoto doesn't currently _need_ service accounts for any good
reason, they enhance the identity demo.

All services now have dedicated accounts except the vote-bot, which continues
to use the "default" account.


klingerf:
Changes to the emojivoto yaml files need to happen in the emojivito repo first. Mind opening up a PR there?

PR #216, opened 2019-03-08
Modifying Makefile and updating README Hugo development instructions

scottcarol:
This PR adds `lint` and `check` targets to the linkerd.io `Makefile` and updates the `README` instructions for building with Hugo.

### DIFF
--- a/linkerd.io/Makefile
+++ b/linkerd.io/Makefile
@@ -1,9 +1,3 @@
-check:
-	htmltest
-
-lint:
-	markdownlint -c .markdownlint.yaml ./content
-
 production-build:
 	hugo --minify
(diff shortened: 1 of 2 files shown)


linkerd.io/Makefile
```diff
@@ -1,3 +1,9 @@
+check:
```
grampelberg: I would rather that the root makefile is used for this, so we keep it DRY. I'd update the root README.md to `hugo serve -s linkerd.io` and have all the commands run from the root of `website`.

scottcarol: Got it, thanks. Does the new README.md text make sense?

PR #243, opened 2019-04-07
add basic design principles doc

wmorgan:


wmorgan:
Personally I think it makes sense at the top level, at least for now

PR #265, opened 2019-04-17
Update books.md

andreasottosson:
Wrong namespace, should be booksapp not default

### DIFF
--- a/linkerd.io/content/2/tasks/books.md
+++ b/linkerd.io/content/2/tasks/books.md
@@ -421,7 +421,7 @@ vary depending on the characteristics of your cluster. To edit the `books`
 service profile, run:
 
 ```bash
-kubectl -n booksapp edit sp/authors.default.svc.cluster.local
+kubectl -n booksapp edit sp/authors.booksapp.svc.cluster.local
 ```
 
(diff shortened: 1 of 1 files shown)


grampelberg:
Would you mind getting DCO to pass? https://github.com/linkerd/website/pull/265/checks?check_run_id=103944090

andreasottosson:
#271 does the same thing, should we close this PR instead?

andreasottosson:
I would also just like to ask if there are more errors in that last chapter of the "Books demo" page, the "Timeouts" one. Read through it and see what you think, it didn't really make sense to me. For example, just above my proposed change it reads "To edit the **books** service profile, run: ..." and then we edit the **authors** service. This was my first time trying linkerd so I might be wrong here =)

grampelberg:
That's a great question, I'll need to do a run through, it has been awhile for me =)

klingerf:
@andreasottosson Really sorry for the delay on this PR, and thanks for your feedback on the books app demo. I went through the timeouts section as you suggested, and it was indeed out of whack. I've submitted #319 to address that, and I'll go ahead and close out this PR.

PR #268, opened 2019-04-17
update docs to reflect that donut emoji is now the breaking emoji

wmorgan:


wmorgan:
These changes are already in.

PR #291, opened 2019-05-01
Update Data Plane Upgrading section given things are injected now by the proxy injector by default

alpeb:
There's a large `jq` command in there. If anyone knows a more succinct way, I'm all ears.


ihcsim:
Superseded by #418.

PR #302, opened 2019-05-13
add icmy-may

PatrickHeneise:


PatrickHeneise:
Done :)

grampelberg:
@PatrickHeneise looks like we're missing DCO, mind signing those commits and this can get merged.

PR #310, opened 2019-05-18
fix up some lint errors

wmorgan:


wmorgan:
AFAICT `htmltest` is disagreeing with hugo's  auto-generated TOC here. Publishing to this website is blocked because two computer programs can't agree on what a valid anchor link looks like.

PR #317, opened 2019-05-26
Fix a couple glaring typos in opening paragraphs

carehart:


carehart:
Thanks for that, Kevin. Here’s the thing, though. I’m using the github web UI. I thought doing so would presume to make it duly signed. I gather now it is not. I have just found an option to solve this, going forward, in a Chrome extension that would add it to future commits. 

As for this one, I have just added the verbiage as a comment on my one commit. Can you tell if that will suffice?  Or if not, is there a way you may be able to override this requirement, since it’s so clearly just about correcting typos?  Not like there can be legal ramifications of me claiming copyright to that! 

But I realize your hands may be tied. If so, I could delete and re-submit the commit and PR anew, if we must. Let me know what you think.

/charlie

wmorgan:
We don't really need the DCO for a change this trivial so I am going to merge it as is. Thanks @carehart!

carehart:
Great. Thanks so much. And I’ll have this covered for future commits, here or elsewhere. Until again.

PR #320, opened 2019-05-29
Add golden metrics description

### LINKED ISSUE #296: Expand the Telemetry and Monitoring with golden metrics description
Expand the Telemetry and Monitoring page with an introductory description of the golden metrics, something like so:

## Success Rate
This is the percentage of successful requests during a time window (1 minute by default).
`linkerd routes` with the optional flag `--o wide` splits this metric into EFFECTIVE_SUCCESS and ACTUAL_SUCCESS. For routes configured with retries, the former calculates the percentage of success after retries (as perceived by the client-side), and the latter before retries (which can expose potential problems with the service).

## Traffic (Requests Per Second)
This gives an overview of how much demand is placed on the service/route.
As with success rates, `linkerd routes` splits this metric into EFFECTIVE_RPS and ACTUAL_RPS, corresponding to rates after and before retries respectively.

## Latencies
Times taken to service requests per service/route are split into 50%, 95% and 99% percentiles.
Lower percentiles give you an overview of the average performance of the system, while tail percentiles help catch outlier behavior.

alpeb:
Fixes #296

### DIFF
--- a/linkerd.io/content/2/features/telemetry.md
+++ b/linkerd.io/content/2/features/telemetry.md
@@ -33,6 +33,29 @@ This data can be consumed in several ways:
   [pre-built Grafana dashboards](/2/features/dashboard/#grafana).
 * Directly from Linkerd's built-in Prometheus instance
 
+## Golden metrics
+
+### Success Rate
+This is the percentage of successful requests during a time window (1 minute by
+default).
+
+In the output of the command `linkerd routes --o wide`, this metric is split
+into EFFECTIVE_SUCCESS and ACTUAL_SUCCESS. For routes configured with retries,
+the former calculates the percentage of success after retries (as perceived by
+the client-side), and the latter before retries (which can expose potential
+problems with the service).
+
+### Traffic (Requests Per Second)
+This gives an overview of how much demand is placed on the service/route. As
+with success rates, `linkerd routes --o wide` splits this metric into
+EFFECTIVE_RPS and ACTUAL_RPS, corresponding to rates after and before retries
+respectively.
+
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/features/telemetry.md
```diff
@@ -33,6 +33,32 @@ This data can be consumed in several ways:
...
+### Traffic (Requests Per Second)
+
+This gives an overview of how much demand is placed on the service/route. As
+with success rates, `linkerd routes --o wide` splits this metric into
+EFFECTIVE_RPS and ACTUAL_RPS, corresponding to rates after and before retries
+respectively.
+
+### Latencies
+
+Times taken to service requests per service/route are split into 50%, 95% and
```
ihcsim: _50th, 95th and 99th percentiles_

linkerd.io/content/2/features/telemetry.md
```diff
@@ -33,6 +33,32 @@ This data can be consumed in several ways:
   [pre-built Grafana dashboards](/2/features/dashboard/#grafana).
 * Directly from Linkerd's built-in Prometheus instance
 
+## Golden metrics
+
+### Success Rate
```
ihcsim: _Success_ as in any responses that aren't 4xx and 5xx? What about gRPC?

alpeb: The rate of success maps to the `classification=failure|success` metric that the proxy generates. I'm assuming success for HTTP 1 requests are those with status 200, and for gRPC requests those with `grpc_status` 0.

Issue #333, opened 2019-06-06
Update markdown-lint-cli to 0.14.1 to address js-yaml issue


cpretzer:
Github has notified us that there is a security issue with js-yaml which is a dependency of markdown-lint-cli.

The fix for markdown-lint-cli is in [v0.14.1](https://github.com/igorshubovych/markdownlint-cli/issues/54).

The fix is to update package.json to use markdown-lint-cli 0.14.1

PR #335, opened 2019-06-07
Upgrade vulnerable sub-dependency: js-yaml

cpretzer:

Subject: Upgrade vulnerable dependency

Problem: js-yaml, a dependency of markdown-lint-cli has a security vulnerability

Solution: Upgrade markdown-lint-cli in package.json to 0.14.1 to upgrade js-yaml to 3.13.1

Fixes: #333

### DIFF
--- a/linkerd.io/content/1/_index.md
+++ b/linkerd.io/content/1/_index.md
@@ -1,3 +1,6 @@
+---
+title: "Overview"
+---
 <!-- markdownlint-disable -->
 <meta http-equiv="Refresh" content="0; url=overview/">
 <!-- markdownlint-enable -->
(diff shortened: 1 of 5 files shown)


cpretzer:
thanks @grampelberg I'm building the new image now

cpretzer:
@scottcarol yep, there were a few links that I cleaned up for the blog post.

The changes to _index.md resolve some failures from *htmltest*. There was no doctype in the generated HTML files.

After doing some digging on the markdown-html generator, it turns out that adding any markup to the .MD file ensures that the *<!doctype html>* tag is added to the generated HTML.
Admittedly, this is a bit of a hack and I'd like to learn the proper way to ensure that an HTML file generated from markdown has the <!doctype> tag.

grampelberg:
@cpretzer I assume that's because of how hugo's doing the rendering path. Happy to walk you through how that's all plumbed together.

PR #351, opened 2019-06-18
Update download banner

dadjeibaah:
Before:
Screen Shot 2019-06-18 at 10 14 44 AM

After:
Screen Shot 2019-06-18 at 10 46 30 AM

### DIFF
--- a/linkerd.io/content/_index.md
+++ b/linkerd.io/content/_index.md
@@ -13,7 +13,7 @@ top_hero:
     url: /community
   image_on_the_right: false
 announcement_banner:
-  description: 2.2.1 is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
+  description: 2.3.2 is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
 description: Linkerd is a transparent proxy that adds service discovery, routing,
   failure handling, and visibility to modern software applications
 keywords:


linkerd.io/content/_index.md
```diff
@@ -13,7 +13,7 @@ top_hero:
     url: /community
   image_on_the_right: false
 announcement_banner:
-  description: 2.2.1 is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
+  description: 2.3.2 is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
```
klingerf: Hmm, we're going to have to remember to bump this with every stable release. If we instead use `L5D2_STABLE_VERSION` here, that should ensure that we're not hardcoding versions (with the caveat that this will print as "stable-2.3.2 is out now!").

If you make this change:

```diff
diff --git a/linkerd.io/content/_index.md b/linkerd.io/content/_index.md
index 9927f30..783f1fb 100644
--- a/linkerd.io/content/_index.md
+++ b/linkerd.io/content/_index.md
@@ -13,7 +13,7 @@ top_hero:
     url: /community
   image_on_the_right: false
 announcement_banner:
-  description: 2.3.2 is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
+  description: L5D2_STABLE_VERSION is out now! Get the newest version. [Download now >](https://github.com/linkerd/linkerd2/releases/)
 description: Linkerd is a transparent proxy that adds service discovery, routing,
   failure handling, and visibility to modern software applications
 keywords:
```

Then that will render as:

image

dadjeibaah: Thanks @klingerf I wasn't sure if there was additional work that needed to be done for the site to pick up that variable. Thanks for testing it out and for the suggestion!

PR #354, opened 2019-06-18
Replace l5d version site params with the release version variable

dadjeibaah:
Every time the website is built for deployment, we run a `sed` replace on a subset of pages to fill in the latest edge and stable versions for Linkerd2. There are other pages that instead, use a site param from the  `config.toml` in the website repo.

This PR aims at making things more consistent. Instead of relying on the site param in the `config.toml`, we just replace every instance of where a site param is used with L5D2_RELEASE_VERSION or L5D2_EDGE_VERSION. This makes `sed` solely responsible for filling in version information when the website is built.


grampelberg:
@dadjeibaah I think this broke something else, take a look at the lint failures (you can run with `make lint` locally).

dadjeibaah:
@grampelberg those lint errors seem unrelated to this change. When running this on the master branch, I still get the same errors.

grampelberg:
Hmm, master is passing without those issues for me, what version are you running?

grampelberg:
Ignore me, something else is the deal.

linkerd.io/layouts/shortcodes/latestversion.html
```diff
@@ -1 +1 @@
-{{ if eq .Page.Section "1" }}{{ site.Params.latest_release_version }}{{ else }}{{ site.Params.l5d2_release_version }}{{ end }}{{- "" -}}
+{{ if eq .Page.Section "1" }}{{ site.Params.latest_release_version }}{{ else }} L5D2_STABLE_VERSION {{ end }}{{- "" -}}
```
klingerf: We should remove the extra spaces that you're adding here. Otherwise the version will look weird with fixed-width text:

image

grampelberg:
@dadjeibaah mind rebasing this on master so that CI passes?

grampelberg:
@dadjeibaah is there any reason we didn’t merge this? Seems like a great PR =)

Issue #355, opened 2019-06-19
Engineers Love Linkerd tweets go off the page


06kellyjac:
In Firefox on a 720p or 1080p screen the tweets go off the screen. At Tablet or Mobile scale it looks fine.

1080p:
![1080p](https://user-images.githubusercontent.com/9866621/59756006-d9db2e00-9280-11e9-98b6-d3b66bc247b6.png)

Tablet:
![Tablet](https://user-images.githubusercontent.com/9866621/59756100-fecfa100-9280-11e9-81d0-89c9e825b073.png)

Phone:
![Phone](https://user-images.githubusercontent.com/9866621/59756151-14dd6180-9281-11e9-9ed7-e2b8cb27ee16.png)

06kellyjac:
I've made an attempt at fixing it:
![Attempted fix](https://user-images.githubusercontent.com/9866621/59756478-b2d12c00-9281-11e9-961e-0ae3fd146ac9.png)

However I think I prefer this look which I had when turning off all the `flex: none` css however that did break the mobile scaling and probably a bunch of other stuff
![Other](https://user-images.githubusercontent.com/9866621/59756532-ced4cd80-9281-11e9-9009-6367477e37dd.png)

PR #356, opened 2019-06-19
Prevented the tweets from overflowing off the page

06kellyjac:
An attempt at fixing issue #355 

Before (1080p):
Before

After (1080p):
After


grampelberg:
@06kellyjac would you mind taking care of DCO? I can merge after that =)

PR #369, opened 2019-06-21
Remove broken JS snippet

klingerf:
I noticed the following JS error on linkerd.io:

Screen Shot 2019-06-21 at 3 26 30 PM

This branch removes the JS that is causing that error, since I couldn't figure out any way to fix it. My guess is that's it's leftover from the previous design. But happy to discard this if there's a better fix.


klingerf:
This issue was fixed by #378; closing.

PR #378, opened 2019-06-25
fix: fix getting started page

### LINKED ISSUE #377: Formatting errors on linkerd1 getting started page
It looks like the CSS for https://linkerd.io/1/getting-started/ is broken. That page now looks like this:

image

dmitry-zaets:
Closes #377


grampelberg:
@dmitry-zaets I would have expected the options to be there as well. I suspect that this used to use the choose your own adventure snippet.

PR #392, opened 2019-07-02
Add PSP doc

### LINKED ISSUE #94: Add minimal PSP policy
- [x] Add minimum possible PSP to security section (see #100)

The minimum PSP for linkerd:

```yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: default
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default'
    apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
    seccomp.security.alpha.kubernetes.io/defaultProfileName:  'docker/default'
    apparmor.security.beta.kubernetes.io/defaultProfileName:  'runtime/default'
spec:
  privileged: false
  # Required to prevent escalations to root.
  allowPrivilegeEscalation: false
  allowedCapabilities:
    # The initContainer for LD2 requires this capability to operate. It is
    # unsafe and *should* be removed.
    - NET_ADMIN
    - NET_RAW
  requiredDropCapabilities:
    - SETPCAP
    - MKNOD
    - AUDIT_WRITE
    - CHOWN
    - DAC_OVERRIDE
    - FOWNER
    - FSETID
    - KILL
    - SETGID
    - SETUID
    - NET_BIND_SERVICE
    - SYS_CHROOT
    - SETFCAP
  # Allow core volume types.
  volumes:
    - 'configMap'
    - 'emptyDir'
    - 'projected'
    - 'secret'
    - 'downwardAPI'
    # Assume that persistentVolumes set up by the cluster admin are safe to use.
    - 'persistentVolumeClaim'
  hostNetwork: false
  hostIPC: false
  hostPID: false
  runAsUser:
    rule: 'RunAsAny'
    # The initContainer for LD2 requires running as root to install the
    # iptables rules. This *should* be set.
    # rule: 'MustRunAsNonRoot'
  seLinux:
    # This policy assumes the nodes are using AppArmor rather than SELinux.
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'MustRunAs'
    ranges:
      # Forbid adding the root group.
      - min: 1
        max: 65535
  fsGroup:
    rule: 'MustRunAs'
    ranges:
      # Forbid adding the root group.
      - min: 1
        max: 65535
  readOnlyRootFilesystem: false
```

ihcsim:
Fixes #94.

### DIFF
--- a/linkerd.io/content/2/tasks/using-psp.md
+++ b/linkerd.io/content/2/tasks/using-psp.md
@@ -19,13 +19,18 @@ Adjust the value of the above label to match your control plane's namespace.
 Notice that to minimize attack surface, all Linux capabilities are dropped from
 the control plane's Pod Security Policy, with the exception of the `NET_ADMIN`
 and `NET_RAW` capabilities. These capabilties provide the `proxy-init` init
-container with runtime privilege to rewrite the pod's iptable. Note that adding
+container with runtime privilege to rewrite the pod's `iptable`. Note that adding
 these capabilities to the Pod Security Policy doesn't make the container a
 [`privileged`](https://kubernetes.io/docs/concepts/workloads/pods/pod/#privileged-mode-for-pod-containers)
 container. The control plane's Pod Security Policy prevents container privilege
 escalation with the `allowPrivilegeEscalation: false` policy. To understand the
 full implication of the `NET_ADMIN` and `NET_RAW` capabilities, refer to the
-Linux capabilities [manual](http://man7.org/linux/man-pages/man7/capabilities.7.html).
+Linux capabilities [manual](http://man7.org/linux/man-pages/man7/capabilities.7.html)
+.
+
+More information on the `iptables` rules used by the `proxy-init` init
+container can be found on the [Architecture](/2/reference/architecture/#linkerd-init)
+page.
 
 If your environment disallow the operation of containers with escalated Linux
 capabilities, Linkerd can be installed with its [CNI plugin](/2/features/cni/),
--- a/linkerd.io/content/2/tasks/using-psp.md
+++ b/linkerd.io/content/2/tasks/using-psp.md
@@ -3,19 +3,31 @@ title = "Using PSP"
 description = "Using Linkerd with pod security policy"
 +++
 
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/tasks/using-psp.md
```diff
@@ -0,0 +1,72 @@
...
+because the policies will vary depending on the security requirements of your
+application. The security context requirement for the Linkerd proxy sidecar
+container will be very similar to that defined in the control plane's Pod
+Security Policy.
+
+For example, the following Pod Security Policy will work with the injected
+`emojivoto` demo application:
+
+```yaml
+cat <<EOF|k apply -f -
```
grampelberg: Could this just be the YAML instead of a command?

linkerd.io/content/2/tasks/using-psp.md
```diff
@@ -0,0 +1,72 @@
...
+
+To view the definition of the control plane's Pod Security Policy, run:
+
+```bash
+kubectl describe psp -l linkerd.io/control-plane-ns=linkerd
+```
+
+Adjust the value of the above label to match your control plane's namespace.
+
+Note that in the CNI setup, the `NET_ADMIN` and `NET_RAW` capabilities are
```
grampelberg: I’d like this to be the meat of the article, it is what users will actually need to use.

linkerd.io/content/2/tasks/using-psp.md
```diff
@@ -0,0 +1,72 @@
...
+  volumes:
+  - configMap
+  - emptyDir
+  - projected
+  - secret
+  - downwardAPI
+  - persistentVolumeClaim
+EOF
+```
+
```
grampelberg: Can we show an example binding, service account and workload for these?

ihcsim:
I am a bit hesitant to include any Linux kernal content in our doc. In the k8s [PSP doc](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#capabilities), they provided a link to the Linux capabilities `man` page. Will adding a link work for us too?

grampelberg:
@ihcsim I think it is still important to say “this is what NET_ADMIN does”, “this is why we need it”, “this is what you can do otherwise”. Linking to a deeper dive on what it does and how it works is great, but we should have some kind of summary at least.

PR #396, opened 2019-07-04
2.4 - Upgrade notes

### LINKED ISSUE #349: Add 2.4 upgrade notes
Use this task to capture the 2.4 upgrade notes.

### Tap RBAC bug

- [x] Done

If the users encountered:
```
The ClusterRoleBinding "linkerd-linkerd-tap" is invalid: roleRef: Invalid value: rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole", Name:"linkerd-linkerd-tap"}: cannot change roleRef
```
they will need to manually delete the `linkerd-linkerd-tap` cluster role binding, and re-run upgrade again. They will run into this issue only if they have previously installed edge-19.5.3 on the same cluster. This bug was fixed in edge-19.5.4.

### Webhook Side-effect

- [x] Done

For k8s pre-1.12, upgrade has to be performed with the `--omit-webhook-side-effects` flag, because the `sideeffect` property of the MWC and VWC is only supported by k8s 1.12+.

### Upgrade check

- [x] Done

If the user runs`linkerd check` against an older control plane, the check will fail complaining that the cluster roles don't exist. This is ok because the `check` command now relies on the label to determine if the RBAC resources exist.
```
linkerd version
Client version: edge-19.6.4
Server version: stable-2.3.2
➜  linkerd2 git:(master) ✗ linkerd check
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
√ is running the minimum Kubernetes API version
√ is running the minimum kubectl version

linkerd-config
--------------
√ control plane Namespace exists
× control plane ClusterRoles exist
    missing ClusterRoles: linkerd-linkerd-controller, linkerd-linkerd-identity, linkerd-linkerd-prometheus, linkerd-linkerd-proxy-injector, linkerd-linkerd-sp-validator, linkerd-linkerd-tap
    see https://linkerd.io/checks/#l5d-existence-cr for hints
```
The `check` will succeed after upgrade is completed.

### HA Setup

- [x] Done

The MWC and VWC will be configured with `FailurePolicy` to `fail`. Make sure that there are more than one proxy injector and sp validator.

### From Manifest

- [x] Done

Make sure this still works with the new proxy injector and sp validator secrets

ihcsim:
Tested the following scenarios with stable-2.3 and stable-2.3.2:

- [x] HA mode - ensures that multiple replicas of the proxy injector and sp validators are provisioned, and work without conflicts, and the MWC and VWC `failurePolicy` is set to `fail`.
- [x] From manifest - ensures that `linkerd upgrade --from-manifest` is compatible with the new `linkerd-proxy-injector-tls` and `linkerd-sp-validator-tls` secrets. The webhooks now read their TLS cert and private keys from these secrets to support multiple replicas, instead of storing them in memory.
- [x] Multi-stage upgrade
- [x] Make sure the control plane's PSP and its RBAC are added during upgrade
- [x] Tested with k8s 1.11 to make sure `linkerd upgrade --omit-webhook-side-effects` doesn't set the MWC and VWC `sideEffects` field
- [x] Make sure all resources are labelled and we can select them with

```sh
`kubectl get all,sa,cm,secret,role,rolebinding,clusterrole,clusterrolebinding,mutatingwebhookconfiguration,validatingwebhookconfiguration,crd,psp --all-namespaces -l linkerd.io/control-plane-ns=linkerd -L linkerd.io/control-plane-ns`
```

Fixes #349.

### DIFF
--- a/linkerd.io/content/2/tasks/upgrade.md
+++ b/linkerd.io/content/2/tasks/upgrade.md
@@ -16,6 +16,82 @@ There are three components that need to be upgraded:
 In this guide, we'll walk you through how to upgrade all three components
 incrementally without taking down any of your services.
 
+## Upgrade notice: stable-2.4.0
+
+Note that the minimum supported Kubernetes version for the `stable-2.4.0`
+release is 1.12.
+
+### Upgrading from stable-2.3.x, edge-19.5.x, edge-19.6.x
+
+Use the `linkerd upgrade` command to upgrade the control plane. This command
+ensures that all existing control plane's configuration and mTLS secret are
+retained.
+
+```bash
+# get the latest stable CLI
+curl -sL https://run.linkerd.io/install | sh
+```
+
+For Kubernetes 1.12+:
+
+```bash
+linkerd upgrade | kubectl apply -f -
+```
+
+For Kubernetes pre-1.12 where the mutating and validating webhook
+configurations' `sideEffects` fields aren't supported:
+
+```bash
+linkerd upgrade --omit-webhook-side-effects | kubectl apply -f -
+```
+
+The `sideEffects` field is added to the Linkerd webhook configurations to
+indicate that the webhooks have no side effects on other resources.
+
+For HA setup, the `linkerd upgrade` command will also retain all previous HA
+configuration. Note that the mutating and validating webhook configuration are
+updated to set its `failurePolicy` field to `fail` to ensure that un-injected
+workloads (as a result of unexpected errors) are rejected during the admission
+process. The HA mode has also been updated to schedule multiple replicas of the
+`linkerd-proxy-injector` and `linkerd-sp-validator` deployments.
+
+For users upgrading from the `edge-19.5.3` release, note that the upgrade
+process will fail with the following error message, due to a naming bug:
+
+```bash
+The ClusterRoleBinding "linkerd-linkerd-tap" is invalid: roleRef: Invalid value:
+rbac.RoleRef{APIGroup:"rbac.authorization.k8s.io", Kind:"ClusterRole",
+Name:"linkerd-linkerd-tap"}: cannot change roleRef
+```
+
+This can be resolved by simply deleting the `linkerd-linkerd-tap` cluster role
+binding resource, and re-run the `linkerd upgrade` command.
+
+For upgrading a multi-stage installation setup, follow the instructions at
+[Upgrading a multi-stage install](/2/tasks/upgrade/#upgading-a-multi-stage-install).
+
+Users who have previously saved the Linkerd control plane's configuration to
+files can follow the instructions at
+[Upgrading via manifests](/2/tasks/upgrade/#upgrading-via-manifests)
+to ensure those configuration are retained by the `linkerd upgrade` command.
+
+Once the `upgrade` command exits, use the `linkerd check` command to confirm
+the control plane is ready. (Note that the `stable-2.4` `linkerd check` command
+doesn't work with older versions of the control plane, because the command is
+updated to select the resources by labels, instead of by names. The benign error
+returned by the command regarding missing required RBAC resources will be
+resolved once the control plane is upgraded to `stable-2.4`.)
+
+When ready, proceed to upgrading the data plane by following the instructions at
+[Upgrade the data plane](#upgrade-the-data-plane).
+
+### Upgrading from stable-2.2.x
+
+Follow the [stable-2.3.0 upgrade instructions](/2/tasks/upgrade/#upgrading-from-stable-2-2-x-1)
+to upgrade the control plane to the stable-2.3.2 release first. Then follow
+[these instructions](/2/tasks/upgrade/#upgrading-from-stable-2-3-x-edge-19-5-x-edge-19-6-x)
+to upgrade the stable-2.3.2 control plane to `stable-2.4.0`.
+
 ## Upgrade notice: stable-2.3.0
 
 `stable-2.3.0` introduces a new `upgrade` command. This command only works for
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -16,6 +16,82 @@ There are three components that need to be upgraded:
...
+For upgrading a multi-stage installation setup, follow the instructions at
+[Upgrading a multi-stage install](/2/tasks/upgrade/#upgading-a-multi-stage-install).
+
+Users who have previously saved the Linkerd control plane's configuration to
+files can follow the instructions at
+[Upgrading via manifests](/2/tasks/upgrade/#upgrading-via-manifests)
+to ensure those configuration are retained by the `linkerd upgrade` command.
+
+Once the `upgrade` command exits, use the `linkerd check` command to confirm
+the control plane is ready. (Note that the `stable-2.4` `linkerd check` command
```
siggy: i don't think we need to dig too far into this why this failure is happening, instead we can just show them what we expect they'll see...

```
{{< note >}}
The `stable-2.4` `linkerd check` command will return an error when run against an older control plane. This error is benign and will resolve itself once the control plane is upgraded to `stable-2.4`:

``(tick)
linkerd-config
--------------
√ control plane Namespace exists
× control plane ClusterRoles exist
    missing ClusterRoles: linkerd-linkerd-controller, linkerd-linkerd-identity, linkerd-linkerd-prometheus, linkerd-linkerd-proxy-injector, linkerd-linkerd-sp-validator, linkerd-linkerd-tap
    see https://linkerd.io/checks/#l5d-existence-cr for hints
``(tick)

{{< /note >}}
```

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -16,6 +16,82 @@ There are three components that need to be upgraded:
...
+returned by the command regarding missing required RBAC resources will be
+resolved once the control plane is upgraded to `stable-2.4`.)
+
+When ready, proceed to upgrading the data plane by following the instructions at
+[Upgrade the data plane](#upgrade-the-data-plane).
+
+### Upgrading from stable-2.2.x
+
+Follow the [stable-2.3.0 upgrade instructions](/2/tasks/upgrade/#upgrading-from-stable-2-2-x-1)
+to upgrade the control plane to the stable-2.3.2 release first. Then follow
```
siggy: curious: are we intentionally going to not support skip upgrades? /cc @grampelberg

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -16,6 +16,82 @@ There are three components that need to be upgraded:
 In this guide, we'll walk you through how to upgrade all three components
 incrementally without taking down any of your services.
 
+## Upgrade notice: stable-2.4.0
+
+Note that the minimum supported Kubernetes version for the `stable-2.4.0`
+release is 1.12.
+
+### Upgrading from stable-2.3.x, edge-19.5.x, edge-19.6.x
```
alpeb: Given that `edge-19.4.5` was released after `stable-2.3` and that yesterday `edge-19.7.1` landed, shouldn't those two be included as well?

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -284,11 +360,36 @@ your Kubernetes cluster, it is time to upgrade the data plane. This will change
...
 your service.
 
-For each of your meshed services, you can re-inject your applications in-place.
-Retrieve your YAML resources via `kubectl`, and pass them through
-`linkerd inject`. This will update the pod spec to have the latest version of
-the `linkerd-proxy` sidecar container. By using `kubectl apply`, Kubernetes will
-do a rolling deploy of your service and update the running pods to the latest
+If your workloads are annotated with the auto-inject
+`linkerd.io/inject: enabled` annotation, then you can use the
+`kubectl set image` command to upgrade the Linkerd proxies. Kubernetes will
```
siggy: cool!

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -284,11 +360,36 @@ your Kubernetes cluster, it is time to upgrade the data plane. This will change
...
+kubectl -n emojivoto set image po linkerd-proxy=gcr.io/linkerd-io/proxy:stable-2.4.0 --all
+```
+
+Note that for Kubernetes 1.15+, you can use the `kubectl rollout restart`
+command to restart all your meshed services.
+
+As the new pods are being created, the proxy injector will auto-inject the new
+version of the proxy into the pods.
+
+If the auto-injection isn't part of your workflow, you can still manually
```
siggy: `If auto-injection is not`

linkerd.io/content/2/tasks/upgrade.md
ihcsim: Yeah, from what I saw on gke 1.13, it didn't even take down the pods.

linkerd.io/content/2/tasks/upgrade.md
ihcsim: I am not sure what's the right approach for this is. From 2.2 to 2.3, they will need to manually delete the old `ca` service, and use the `linkerd install --ignore-cluster <all-previously-provided-flags> | kubectl apply -f -` approach. I thought if things go wrong, it will be easier to debug and reason about, if users upgrade to 2.3 first especially, since that upgrade path is very well tested.

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -16,6 +16,82 @@ There are three components that need to be upgraded:
...
+curl -sL https://run.linkerd.io/install | sh
+```
+
+For Kubernetes 1.12+:
+
+```bash
+linkerd upgrade | kubectl apply -f -
+```
+
+For Kubernetes pre-1.12 where the mutating and validating webhook
```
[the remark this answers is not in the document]
ihcsim: Yeah, I debated on this one too. Without this flag, upgrade will fail on k8s pre-1.12 clusters. I think it's fair (at least for this release) to show users who are still on older k8s clusters how to move forward with the upgrade. cc @grampelberg

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -16,6 +16,97 @@ There are three components that need to be upgraded:
...
+```
+
+{{< /note >}}
+
+When ready, proceed to upgrading the data plane by following the instructions at
+[Upgrade the data plane](#upgrade-the-data-plane).
+
+### Upgrading from stable-2.2.x
+
+Follow the [stable-2.3.0 upgrade instructions](/2/tasks/upgrade/#upgrading-from-stable-2-2-x-1)
```
alpeb: I believe the `-1` at the end of that URL should not be there

linkerd.io/content/2/tasks/upgrade.md
alpeb: Will this work when upgrading data planes from 2.3? If the pods won't get recreated then the webhook won't get called, and the new env vars for identity won't get set. IIUC, only the container image will be changed.

alpeb: Ok, identity was actually set on 2.3.0, so my comment would apply for releases before that.

linkerd.io/content/2/tasks/upgrade.md
ihcsim: Actually, it is needed, because now there are two _Upgrading from stable 2.2.x_ anchors in the doc. The `-1` suffix points to the one under the 2.3 upgrade notes.

linkerd.io/content/2/tasks/upgrade.md
ihcsim: Good point. I'm hoping that any pre-2.3.0 control planes will have to be upgraded to 2.3.x first, so that the old `ca` service is removed and the data plane get assigned the proper identity configuration, before the next hop to 2.4. There is a separate comment in this PR on that.

alpeb: Even if the user follows the proper instructions for upgrading the control plane, I didn't see any special instructions for the data plane when coming from pre-2.3.0.
Given the `Upgrade the data plane` section applies to all versions, I think we shouldn't instruct on doing upgrades solely through `set image` as that will miss other parts of the injection mechanism that have changed across versions (identity, default SecurityContext, maybe changes in annotations)

ihcsim: True. I kinda don't want to divide up the data plane upgrade section into more subsections, for different versions of Linkerd. Ultimately, `kubectl rollout restart` is the solution. Until then, can we just say that pods restart is required (so that the proxy injector can inject the newer version of the proxy into the workload), and it's up to the users how they want to restart their pods (helm, `kubectl set`, `kubectl apply`, `kubectl delete` if downtime is acceptable etc.)?

alpeb: Sounds good to me. I've just realized I pushed #291 long ago with a monstrous alternative for there where `kubectl rollout` is not available. TIOLI for your use ;-)

linkerd.io/content/2/tasks/upgrade.md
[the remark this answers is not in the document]
kleimkuhler: I think it is a good idea for this release. I think we should plan to merge as is unless @grampelberg feels otherwise?

linkerd.io/content/2/tasks/upgrade.md
grampelberg: I am apparently missing all the pings, I would like to explicitly not support skip upgrades.

PR #403, opened 2019-07-09
Rollup stable-2.4.0 website changes

kleimkuhler:


kleimkuhler:
Sorry for all the forced pushes on here. A few of these commits had authors and signoffs that did not lineup.

PR #479, opened 2019-08-22
Use sha256sum instead of openssl in the install scripts

joakimr-axis:
It seems sha256sum is more ubiquitous than openssl which would suggest
using it in the install scripts make them more portable. Also using
shell script built-in functionality to trim the sha256sum output which
eliminates the depdency towards awk.

Change-Id: I2e6ceed6ffbb406bc440d5f7d2f816d1f23b4f85
Signed-off-by: Joakim Roubert <joakimr@axis.com>

### DIFF
--- a/run.linkerd.io/public/install
+++ b/run.linkerd.io/public/install
@@ -22,7 +22,8 @@ url="https://github.com/linkerd/linkerd2/releases/download/${LINKERD2_VERSION}/$
   curl -LO "${url}"
   echo ""
   echo "Download complete!, validating checksum..."
-  checksum=$(openssl dgst -sha256 "${filename}" | awk '{ print $2 }')
+  checksum=$(sha256sum "${filename}")
+  checksum=${checksum%%[[:blank:]]*}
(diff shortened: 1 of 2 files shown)


grampelberg:
I guess I merged your first PR before looking through all of them. Is this the final one? Would you mind rebasing so it can merge?

joakimr-axis:
> I guess I merged your first PR before looking through all of them. Is this the final one? Would you mind rebasing so it can merge?

Wow, you're fast! I'm not used to maintainers this active; my idea was I guess a bit unclear but it was to provide some different versions of what I wanted to do. Maintainers in some projects really don't like the use of built-in shell script functionality in shell scripts (I do) and so on, so that's why I did the different ones.

Anyways: yes, this one is actually the one I prefer. I will rebase first thing tomorrow (evening here in my neck of the woods).

Thank you for your swift response!

joakimr-axis:
Conflicts resolved now.

grampelberg:
Thanks! This is great =)

PR #504, opened 2019-09-05
Minor Service Profile doc fix

### LINKED ISSUE #488: SP template and docs do not sync
### What problem are you trying to solve?
When I trying to create service profile from scratch using `linkerd profile -n emjoivoto web-svc --template` it suggested that I used `isFailure: true`, but when I check the [documentation](https://linkerd.io/2/reference/service-profiles/) is only have `isSuccess`
###  should the problem be solved?
Yes, this [code](https://github.com/linkerd/linkerd2/blob/f4608f2bcd3daeb1ad4ba7a7b89d4cc58536226c/pkg/profiles/template.go#L74) should be updated using documentation or the docs it self should be updated.

alpeb:
Fixes #488

Fixed doc for spec.routes.responseClasses.condition.isFailure
(The doc referred `isSuccess`, which doesn't exist)

Signed-off-by: Alejandro Pedraza Borrero <alejandro@buoyant.io>


ihcsim:
Awesome! :+1:

While you are here, can I bother you to include these new properties too?

* Add `spec.dstOverrides`, as seen [here](https://github.com/linkerd/linkerd2/blob/fe0783dff5e3d9ac58cc48af9b6d2ae7e507a7e7/controller/gen/apis/serviceprofile/v1alpha2/types.go#L33).
* Rename `routes.responses` to `routes.responseClasses`
* Add the new `WeightedDst` as seen [here](https://github.com/linkerd/linkerd2/blob/fe0783dff5e3d9ac58cc48af9b6d2ae7e507a7e7/controller/gen/apis/serviceprofile/v1alpha2/types.go#L84-L87).
* Under `Request Match` and `Response Match`, the `must contain exactly one of the following fields:` can be replaced with `may contain the following fields:`, to match what the template generated by `linkerd profile` says; i.e. `# Conditions can be combined using 'all', 'any', and 'not'`.

Thanks.

alpeb:
> * Add `spec.dstOverrides`, as seen [here](https://github.com/linkerd/linkerd2/blob/fe0783dff5e3d9ac58cc48af9b6d2ae7e507a7e7/controller/gen/apis/serviceprofile/v1alpha2/types.go#L33).

Turns out this is meant to be hidden from users, and is only meant to be filled by the API.

> * Rename `routes.responses` to `routes.responseClasses`

:+1: 

> * Add the new `WeightedDst` as seen [here](https://github.com/linkerd/linkerd2/blob/fe0783dff5e3d9ac58cc48af9b6d2ae7e507a7e7/controller/gen/apis/serviceprofile/v1alpha2/types.go#L84-L87).

Same comment as above for `dstOverrides`.

> * Under `Request Match` and `Response Match`, the `must contain exactly one of the following fields:` can be replaced with `may contain the following fields:`, to match what the template generated by `linkerd profile` says; i.e. `# Conditions can be combined using 'all', 'any', and 'not'`.

I believe `must contain exactly one of the following fields` is still more accurate, given the user is really forced to use one of the possibilities only, as shown in the [protobuf definition](https://github.com/linkerd/linkerd2-proxy-api/blob/ddbc3a4f7f8b0058801f896d27974d19ee98094c/proto/destination.proto#L205-L211).
I'll add an example section based on the template to make things clearer.

linkerd.io/content/2/reference/service-profiles.md
```diff
@@ -46,6 +46,43 @@ A request match object must contain _exactly one_ of the following fields:
 | `not` | a [request match](#request-match) object which must _not_ match |
 {{< /table >}}
 
+### Request Match Usage Examples
+
+#### The simplest condition is a path regular expression
```
ihcsim: I don't think we need to mark these as headers; just an ordinary sentence suffixed with `:` should do.

alpeb: Agreed, that didn't look right hehe

linkerd.io/content/2/reference/service-profiles.md
```diff
@@ -46,6 +46,42 @@ A request match object must contain _exactly one_ of the following fields:
...
+```
+
+This is a condition that checks the request method:
+
+```yaml
+method: POST
+```
+
+If more than one condition field is set, all of them must be satisfied. This is
+equivalent to using the 'all' condition:
```
ihcsim: Sorry, I missed this on the last pass; maybe use backticks for _all_ here, and in line 72. :pray:

alpeb: whoops I just merged...

ihcsim: No worries. Next time. :smile:

PR #547, opened 2019-10-06
2.6 upgrade notes

ihcsim:
This PR includes the 2.6 upgrade notes. I also re-organize the upgrade docs to re-use common sections like _upgrading the cli_, _upgrading the control plane_ and _upgrading the data plane_, so that we won't have to repeat those commands in subsequent upgrade notes.

I specifically call out the data plane downtime, due to the migration of the `destination` container into its own workload. We can discuss in-person to see if there are ways around it.


ihcsim:
@cpretzer The 503 errors didn't show up in the proxy logs. But you will be able to reproduce it using a before and after upgrade injected `curl` pod. In my case, I just curl'ed `web-svc.emojivoto/api/list`.

PR #548, opened 2019-10-06
Document the nginx.ingress.kubernetes.io/upstream-vhost annotation

cpretzer:
Documenting the `nginx.ingress.kubernetes.io/upstream-vhost` annotation with a static value for now.

We'll need to explore the correct way to dynamically set this value. the `$` nginx variables are not available to the annotation based on the testing that I've done. Will have to dig in to where this is added in the nginx configuration

### DIFF
--- a/linkerd.io/content/2/tasks/using-ingress.md
+++ b/linkerd.io/content/2/tasks/using-ingress.md
@@ -79,6 +79,7 @@ The important annotation here for http traffic is:
 
 When using gRPC, NGINX has a distinct set of directives for managing
  headers:
+
 ```yaml
     nginx.ingress.kubernetes.io/configuration-snippet: |
       grpc_set_header l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;
@@ -87,7 +88,6 @@ When using gRPC, NGINX has a distinct set of directives for managing
 In the examples above and below, we include both headers for brevity. You
 should include the directives that are relevant for the type of traffic that
 your service uses.
-
 
 In addition, the line below ensures that the `edge` between the ingress
 controller and the backend service are shown in the
--- a/linkerd.io/content/2/tasks/using-ingress.md
+++ b/linkerd.io/content/2/tasks/using-ingress.md
@@ -51,6 +51,7 @@ metadata:
   namespace: emojivoto
   annotations:
     kubernetes.io/ingress.class: "nginx"
+    nginx.ingress.kubernetes.io/upstream-vhost: web-svc.emojivoto.svc.cluster.local:8080
     nginx.ingress.kubernetes.io/configuration-snippet: |
       proxy_set_header l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;
 spec:
@@ -60,7 +61,7 @@ spec:
       paths:
       - backend:
           serviceName: web-svc
-          servicePort: 80
+          servicePort: 8080
 ```
 
 The important annotation here is:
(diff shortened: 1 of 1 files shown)


cpretzer:
Addresses #546

grampelberg:
@cpretzer this seems like a proxy bug? The fact that it works with http/1 and not http/2 makes me wonder what's going on.

grampelberg:
@cpretzer where'd we get on this one?

cpretzer:
@grampelberg we're still waiting to hear back on #546. 

In the meantime, I'm updating this PR to include the `grpc_*` directives that I used to configure the ingress for grpc traffic.

grampelberg:
Tested with `grpc_*` and it is working for me as well. Might as well just suggest having both lines in there and be done with it.

cpretzer:
@grampelberg, I've updated the PR to combine the gRPC directives along with the original `proxy_*` directives, with a note about using the directive that is appropriate for the traffic to the service.

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -51,25 +51,52 @@ metadata:
   namespace: emojivoto
   annotations:
     kubernetes.io/ingress.class: "nginx"
+    nginx.ingress.kubernetes.io/upstream-vhost: web-svc.emojivoto.svc.cluster.local:8080
```
grampelberg: Don't need this anymore.

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -51,25 +51,52 @@ metadata:
...
     nginx.ingress.kubernetes.io/configuration-snippet: |
+
+      # for http traffic
```
grampelberg: In my testing, we can just leave both there all the time and not worry about it. I'd rather just keep it simple (show both lines) and remove the comments.

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -51,25 +51,52 @@ metadata:
...
     http:
       paths:
       - backend:
           serviceName: web-svc
-          servicePort: 80
+          servicePort: 8080
 ```
 
-The important annotation here is:
+The important annotation here for http traffic is:
```
grampelberg: Might as well just smash these together and have a bit about how grpc traffic is treated differently.

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -51,25 +51,52 @@ metadata:
...
+```yaml
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      grpc_set_header l5d-dst-override $service_name.$namespace.svc.cluster.local:$service_port;
+```
+
+In the examples above and below, we include both headers for brevity. You
+should include the directives that are relevant for the type of traffic that
+your service uses.
+
+In addition, the line below ensures that the `edge` between the ingress
```
grampelberg: I'd leave the `edge` conversation out as we just weren't doing things correctly to begin with (and that's how folks found out about it).

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -51,25 +51,52 @@ metadata:
...
+controller and the backend service are shown in the
+[`linkerd edges`](/2/reference/cli/edges/) command:
+
+```yaml
+nginx.ingress.kubernetes.io/upstream-vhost: web-svc.emojivoto.svc.cluster.local:8080
```
grampelberg: Get rid of this one too.

PR #551, opened 2019-10-09
Allow installing parallel linkerd versions and custom install dirs.

memory:
- Add an `INSTALLROOT` var to the install script, defaulting
  to `${HOME}/.linkerd2`

- Save the downloaded binary into `${INSTALLROOT}/bin/linkerd-${LINKERD2_VERSION}`

- Symlink `${INSTALLROOT}/bin/linkerd` to the versioned binary

- Do not re-download binaries if the SHA checksum is passing

- Update documentation to note the presence of the versioned binaries

- This time, it is all pedantically POSIX shell

Signed-off-by: Nathan J. Mehl <n@oden.io>

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -53,9 +53,20 @@ jobs:
       shell: bash
       run: |
         for script in $(find run.linkerd.io/install*); do
+          [ $(uname -s) = "Darwin" ] && statcmd="stat -f %m" || statcmd="stat -c %Y"
           sh ${script}
           set -x
           client=$(~/.linkerd2/bin/linkerd version --client --short)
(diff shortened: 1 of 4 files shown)


memory:
@grampelberg this has now been tested on:

- macOS
    * /bin/bash
    * /bin/sh
    * /bin/sh
- debian
    * /bin/sh
    * /bin/bash
    * /usr/bin/zsh
    * /usr/bin/posh

The notable portability changes vs the first version are:

- remove the `function` keyword (sigh)
- use `sed` instead of infix variable notation to extract the sha hash from openssl/shasum output
- quote the `tr` arguments

Issue #552, opened 2019-10-09
automate install script testing


siggy:
Our `install` and `install-edge` scripts are manually reviewed, but not tested during CI. This can cause issues like #550, where the script may work in some environments/shells but not others.

Update CI to run https://www.shellcheck.net on the scripts, and also execute the script to confirm linkerd is correctly installed.

grampelberg:
TIL, this is great. I really wish there was a way to run OSX containers, doing some automated linux testing will be easy.

grampelberg:
Fixed with #554

PR #554, opened 2019-10-09
Testing actions

grampelberg:


grampelberg:
Fixes #552

.github/workflows/install.yml
```diff
@@ -0,0 +1,61 @@
...
+
+    steps:
+    - uses: actions/download-artifact@v1
+      with:
+        name: run.linkerd.io
+
+    - name: Test install script execution
+      shell: bash
+      run: |
+        for script in $(find run.linkerd.io/install*); do
```
memory: You should probably add a second internal loop here:

```
for a b; do
  (the test)
done
```

The behavior of the second execution in a row of the script should be to validate that the on-disk copy of the cli passes the shasum check and exit cleanly without redownloading. (Maybe check that the mtime on the cli file hasn't changed since before the second check?)

If you _really_ want to test all cases you could add a third test where you break the shasum check (e.g. `echo x >> linkerd_file`) and test that the script correctly susses that out and re-downloads it.

memory: (Oh, hm, all of the above is premised on the idea that my PR has already been merged.  Maybe I'll add it there. :)

PR #561, opened 2019-10-13
Update 'from-manifest' upgrade docs

ihcsim:
This PR updates the `--from-manifests` portion of the upgrade docs, by removing the first example where the whole output of the `linkerd install` command is written to a manifest file. As filed in https://github.com/linkerd/linkerd2/issues/3559, this approach is now causing the `linkerd upgrade` command to fail due to the `k8s.FakeClientSet` not being able to deserialize the `APIService` kind. Since the original intent of this option (see https://github.com/linkerd/linkerd2/issues/2629) is to read the `linkerd-config` config map, I think it's fair to remove the failing example in the docs.

Fixes https://github.com/linkerd/linkerd2/issues/3559.

### DIFF
--- a/linkerd.io/content/2/tasks/upgrade.md
+++ b/linkerd.io/content/2/tasks/upgrade.md
@@ -255,25 +255,13 @@ linkerd upgrade control-plane | kubectl apply --prune -l linkerd.io/control-plan
 
 #### Upgrading via manifests
 
-`edge-19.4.5` introduced a new `--from-manifests` flag to `linkerd upgrade`
-allowing manually feeding a previously saved output of `linkerd install` into
-the command, instead of requiring a connection to the cluster to fetch the
-config:
-
-```bash
-# save Linkerd installation manifest
-linkerd install > linkerd-install.yaml
-
-# deploy Linkerd
-cat linkerd-install.yaml | kubectl apply -f -
-
-# upgrade Linkerd via manifests
-cat linkerd-install.yaml | linkerd upgrade --from-manifests -
-```
-
-Alternatively, if you have already installed Linkerd without saving a manifest,
-you may save the relevant Linkerd resources from your existing installation for
-use in upgrading later.
+By default, the `linkerd upgrade` command reuses the existing `linkerd-config`
+config map and the `linkerd-identity-issuer` secret, by fetching them via the
+the Kubernetes API. `edge-19.4.5` introduced a new `--from-manifests` flag to
+allow the upgrade command to read the `linkerd-config` config map and the
+`linkerd-identity-issuer` secret from a static YAML file. This option is
+relevant to CI/CD workflows where the Linkerd configuration is managed by a
+configuration repository.
 
 ```bash
 kubectl -n linkerd get \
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -255,33 +255,21 @@ linkerd upgrade control-plane | kubectl apply --prune -l linkerd.io/control-plan
...
-Alternatively, if you have already installed Linkerd without saving a manifest,
-you may save the relevant Linkerd resources from your existing installation for
-use in upgrading later.
+By default, the `linkerd upgrade` command reuses the existing `linkerd-config`
+config map and the `linkerd-identity-issuer` secret, by fetching them via the
+the Kubernetes API. `edge-19.4.5` introduced a new `--from-manifests` flag to
+allow the upgrade command to read the `linkerd-config` config map and the
+`linkerd-identity-issuer` secret from a static YAML file. This option is
+relevant to CI/CD workflows where the Linkerd configuration is managed by a
+configuration repository.
```
siggy: Good catch @ihcsim. This has been broken since `edge-19.8.1` / `stable-2.5.0`, specifically from linkerd/linkerd2#3167, where I added the tap APIService to `linkerd install`, but forgot to update fake clientset support.

I've just posted linkerd/linkerd2#3569 to address this oversight.

Of course these docs still need updating because `linkerd install | linkerd upgrade --from-manifests -` won't work between `edge-19.8.1` and `stable-2.6.0`, but going forward I think this is a convenient use case to keep documenting. Consider adding back something like:

For releases prior to `edge-19.8.1` / `stable-2.5.0`, and after `stable-2.6.0`, you may pipe a full linkerd install manifest into the upgrade command:
```
linkerd install > linkerd-install.yaml

# deploy Linkerd
cat linkerd-install.yaml | kubectl apply -f -

# upgrade Linkerd via manifests
cat linkerd-install.yaml | linkerd upgrade --from-manifests -
```

jon-walton:
Hi @ihcsim @siggy 

In my view, the point of `--from-manifests` is to allow air-gapped upgrades. As per comment https://github.com/linkerd/linkerd2/issues/3559#issuecomment-540939495, my install was done via `linkerd install > linkerd-install.yaml`, the yaml saved to a git repo and argocd deploying the yaml.

Each cluster has a separate `linkerd-install.yaml`, containing it's specific config.

I've previously been able to do `linkerd upgrade --from-manifests ./install-prod.yaml > ./install-prod.yaml` and commit the new manifest without the upgrade process touching the cluster. This avoids accidentally pulling in a dev/staging config from the cluster and saving it in the prod manifest, depending on where the kube context happens to be pointing

```bash
kubectl -n linkerd get \
  secret/linkerd-identity-issuer \
  configmap/linkerd-config \
  -oyaml > linkerd-manifests.yaml
```

defeats the purpose of `--from-manifests`, in my view

ihcsim:
@jon-walton Understood. We have a fix for it in https://github.com/linkerd/linkerd2/pull/3569, which is very close to getting merged. The reason why the docs mentioned `kubectl -n linkerd get  secret/linkerd-identity-issuer configmap/linkerd-config -oyaml > linkerd-manifests.yaml` is because those are the resources that `upgrade` pulls from the cluster. But yes, we will definitely continue to support the `linkerd upgrade --from-manifests ./install-prod.yaml > ./install-prod.yaml` use case.

Meanwhile, are you blocked by this? As you mentioned in slack, the `APIService` kind is the problematic one.

PR #568, opened 2019-10-17
Add --name flag to Helm docs for Helm install procedure

scottcarol:
This PR adds `--name=linkerd2` to the Helm install procedure docs so Helm release name is `linkerd2`. Also adds forward slashes to HA installation code so it can be copied as-is.


ihcsim:
Don't worry about the CircleCI errors; we are no longer using CircleCI afaict.

scottcarol:
Adds `--name=linkerd2` to the Helm install procedure docs so Helm release name
is linkerd2. Also adds forward slashes to HA installation code so it can be
copied as-is.

PR #570, opened 2019-10-24
Update edge version to edge-19.10.4 for website

zaharidichev:
Signed-off-by: zaharidichev <zaharidichev@gmail.com>


zaharidichev:
The CI build failed for some reason that I cannot quite explain. This is why we do not have the image published...

PR #579, opened 2019-11-06
Update a-service-mesh-for-kubernetes-part-xi-egress.md

DavidZisky:
Changed "egress" to "ingress" when they were mistaken at the beginning (so it could result in misunderstandings)


grampelberg:
This is great! Check out the [contributing doc](https://github.com/linkerd/website/blob/master/CONTRIBUTING.md#developer-certificate-of-origin) to understand what DCO is and how to get it to pass.

DavidZisky:
As far as I can see "Lint html and check for dead links" fails while checking website which is still available: https://www.techrepublic.com/article/how-twitters-fail-whale-could-save-your-company/

can we just retest or you guys want to take a look?

grampelberg:
Sounds transient, your changes should be live now =)

Issue #587, opened 2019-11-16
Add docs for using linkerd with a private docker repo and air-gapped environment


cpretzer:
Document will include:
- a list of latest stable and edge releases
- instructions for updating YAML files to pull from the private docker repo

PR #589, opened 2019-11-19
#587 add private repo info

cpretzer:

This is a draft at the docs for using a private repo for air-gapped environments.

I'd like to be able to dynamically fill in the latest stable and edge release versions. Are there variables for that?

Also looking for a way to expand on the CI/CD portion by dynamically setting the image names using helm charts or other options.

### DIFF
--- a/linkerd.io/content/2/tasks/using-a-private-docker-repository.md
+++ b/linkerd.io/content/2/tasks/using-a-private-docker-repository.md
@@ -8,10 +8,12 @@ Linkerd images. This scenario requires knowing the names and locations of the
 docker images used by the Linkerd control and data planes so that you can
 store them in your private repository.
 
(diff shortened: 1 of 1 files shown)


grampelberg:
I'm not a huge fan of this one ... it feels obvious and highly specific to folks who are running with airgapped clusters. I might be totally off the mark though?

cpretzer:
@grampelberg I agree that this content is obvious and highly specific.

The part that I think is valuable is the bash one-liner to get the unique images that are output from the `linkerd install` command. It's not entirely obvious to folks that `linkerd install` writes out a bunch of YAML, so this part of the doc will help them to get that information

grampelberg:
@cpretzer should we just merge this? As a standalone doc it is probably fine.

cpretzer:
yep, I'm okay with merging it

cpretzer:
@siggy that's a good idea, thanks. I'll do that in another PR

PR #595, opened 2019-11-27
Define workflow to rotate trust root

zaharidichev:
Defines the steps needed to take in order to rotate trust roots. 

Signed-off-by: zaharidichev <zaharidichev@gmail.com>


linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+## Updating the `linkerd-config` configmap to contain both trust roots
+
+Now that you have a new trust root you need to add it to the bundle in the
+Linkerd config. This will allow the proxies, when re-injected to be configured
+with both roots at the same time. In order to do that, first encode your trust
+root as a single line:
+
+```bash
```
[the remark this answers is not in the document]
zaharidichev: > Use linkerd upgrade to push the updated config map and issuer secret:
> 
> linkerd upgrade --from-manifests=linkerd-config.yaml --identity-issuer-certificate-file=issuer.crt --identity-issuer-key-file=issuer.key 

Its neat that we can do that in one go here. However are there any race implications? For example, imagine the following sequence of unfortunate events that happen after we type this command: 

1. The config map is updated to contain both roots
2. The linkerd-identity-issuer is updated to contain the new certs.
3. The identity service picks up the new certs
4. A proxy with the old roots (still) is due to renew its cert and fires a CSR to the identity service (with the old roots only) and fails...
5. Proxies are rolled with the new trust roots.

This is a total black swan but I thought it was worth calling out. Not event sure whether this is possible in practice. Is it worth splitting this into two steps as: 

1. Ensure the proxies are rolled with the bundled roots
2. Update the linkerd-identity-issuer secret.

@adleong since you outlines the workflow initially, do you think that the scenario described above can lead to problems ?

grampelberg:
@ihcsim can't use `step` to do the bundle as it isn't technically a "bundle".

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,136 @@
...
+
+This will open an editor. You need to append the formatted trust root data to
+the old trust root defined in the `trustAnchorsPem` field, changing it to be in
+the form:
+
+```json
+"trustAnchorsPem":"<old-trust-root><new-trust-root>"
+```
+
+## Reinjecting the proxies with the new trust anchors
```
grampelberg: Can't we just restart the pods to make this work? I'm pretty hesitant to have folks use `inject --manual`.

zaharidichev: I am pretty sure we cannot. The roots are baked into the env vars of the container definitions.

zaharidichev:
@ihcsim @grampelberg  
You know, I just realized that there is a problem with this flow. The problem is that after we update the identitiy-issuer-secret and the identitiy service picks it up, it will try to verify it with the roots that its holding. These roots are being loaded on startup and will not be updated to the roots bundle. At that point the identitiy service will fail. So unfortunatelly we actually need to implement dynamic reloading of the trust anchors for the identitiy service (which will be pretty much watching the config map and updating the trust anchor on change). Was really hoping to not have to do that at that point but I dont see other options.

ihcsim:
> So unfortunatelly we actually need to implement dynamic reloading of the trust anchors for the identitiy service 

Per standup convo, the control plane will need to be restarted as part of this process (using `linkerd upgrade`), to ensure consistency between the trust roots in the config map, the identity's memory and identity's proxy.

zaharidichev:
@ihcsim @grampelberg 

I updated the docs. Adding the anchors flag to the upgrade command certainly makes things look a lot clearner (no `awk`, etc). Now this pr assumes we have the functionality from [here](https://github.com/linkerd/linkerd2/pull/3696) and [here](https://github.com/linkerd/linkerd2/pull/3771) merged. 

Also with respect to providing one liner commands andtrying to minimise user interaction, I have consolidated some things to be more copy&paste friendly. However I still tried to keep it so the user is aware of what is happening at each stage. Let me know what you think.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+rm original-trust.crt
+```
+
+## Deploying the new bundle to Linkerd
+
+You can resort to using `linkerd upgrade` in order to upgrade Linkerd to work
+with the new root bundle:
+
+```bash
+bin/linkerd upgrade  --identity-trust-anchors-file=./bundle.crt | kubectl apply -f -
```
ihcsim: We should use `linkerd` instead of `bin/lnkerd` here.

Also, maybe add a `linkerd check` after `upgrade` and all the control plane pods are restarted.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+````
+
+This will rotate the proxies of the Linkerd components and they will be
+reconfigured with the new root certs. Now it is time rotate the proxy for all
+injected workloads in your cluster. Doing that for the `emojivoto` namespace for
+example would look like:
+
+```bash
+kubectl -n emojivoto get deploy -o yaml | linkerd inject  --manual - | kubectl apply -f -
```
ihcsim: Can we use `kubectl -n emojivoto rollout restart deploy` here, with a note saying if workloads are injected with `linkerd inject --manual`, they will have to re-inject it? I think `rollout restart` should take care of most cases.

Then, run `linkerd check --proxy`.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## Updating the identity issuer certificate
+
+Now that all proxies are updated to include the additional trust root, it is
+safe to rotate the identity issuer certificate by using the `upgrade` command
+again:
+
+```bash
+bin/linkerd upgrade   --identity-issuer-certificate-file=./issuer-new.crt --identity-issuer-key-file=./issuer-new.key | kubectl apply -f -
```
ihcsim: Use `linkerd` instead of `bin/linkerd`.

Then run `linkerd check --proxy`.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+
+```bash
+bin/linkerd upgrade  --identity-trust-anchors-file=./ca-new.crt  | kubectl apply -f -
+```
+
+Additionally you can use the `inject` command to bring the configuration
+of your other injected resources up to date:
+
+```bash
+kubectl -n emojivoto get deploy -o yaml | linkerd inject  --manual - | kubectl apply -f -
```
ihcsim: As above, can we use `rollout restart` here, with a note saying reinject is required for workloads meshed with `linkerd inject --manual`?

Then add a `linkerd check --proxy`.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+9s          Normal   IssuerUpdated   deployment/linkerd-identity   Updated identity issuer
+```
+
+## Removing the old trust root
+
+At this point, it is safe to remove the old bundle and rotate the proxies once
+again. Using the `upgrade` command can do that for the Linkerd components:
+
+```bash
+bin/linkerd upgrade  --identity-trust-anchors-file=./ca-new.crt  | kubectl apply -f -
```
ihcsim: As above, maybe add a `linkerd check` command here.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
zaharidichev: True, I did not realize for a little bit that this will work.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
zaharidichev: You prefer that instead of simply emphasizing in the beggining that check should be ran after each step?

adleong:
I have a lot of thoughts about this workflow and the corresponding changes to check and upgrade and rather than sprinkle them throughout, I thought it would be easier to gather them all in one place.  Here's what I think would be a pretty reasonable user experience based on this workflow:

https://docs.google.com/document/d/148CPU0bUbz2cPFQ8Sj4UDOrU34QdqSDCll3hiXvb_NY/edit?usp=sharing

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
++++
+title = "Rotating the identity trust root"
```
adleong: The goal is the rotate the issuer certificate.  Rotating the trust root is just a necessary step if we do not have a valid root key.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+description = "Follow this work flow to update your identity trust root"
++++
+
+By default, the trust root that Linkerd uses to issue TLS certificates to the
```
adleong: the trust root isn't used to issue certificates, the issuer key and cert are.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+
+...
+```
+
+Running this command at the end of each step and ensuring that all checks pass
+should give you sufficient guarantee that your progressing in the right
+direction.
+
+{{< note >}}
+There is likelihood that this process may incur downtime. Therefore it is
```
adleong: This isn't true as long as the controller is deployed in HA mode.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+step certificate create identity.linkerd.cluster.local ca-new.crt ca-new.key --profile root-ca --no-password --insecure
+```
+
+Note that the above command will generate a root certificate. Alternately, you
+can use your existing trust root to generate an intermediate certificate for
+Linkerd to use as its mTLS trust root. Now generate the intermediate
+certificate and key pair:
+
+```bash
+step certificate create identity.linkerd.cluster.local issuer-new.crt issuer-new.key --ca ca-new.crt --ca-key ca-new.key --profile intermediate-ca --not-after 8760h --no-password --insecure
```
adleong: what does the `--insecure` flag do, other than look really really scary?

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+## Updating the `linkerd-config` configmap to contain both trust roots
```
adleong: this step doesn't actually update the configmap yet

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+```bash
+kubectl -n linkerd get cm linkerd-config -o=jsonpath='{.data.global}' |  \
+jq -r .identityContext.trustAnchorsPem > original-trust.crt &&
+step certificate bundle ca-new.crt original-trust.crt bundle.crt &&
+rm original-trust.crt
+```
+
+## Deploying the new bundle to Linkerd
+
+You can resort to using `linkerd upgrade` in order to upgrade Linkerd to work
```
adleong: the word "resort" makes it sounds like this is a bad thing

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,150 @@
...
+with the new root bundle:
+
+```bash
+bin/linkerd upgrade  --identity-trust-anchors-file=./bundle.crt | kubectl apply -f -
+````
+
+This will rotate the proxies of the Linkerd components and they will be
+reconfigured with the new root certs. Now it is time rotate the proxy for all
```
adleong: I don't think "rotate" is a verb that applies to proxies.  How about "restart"?

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
ihcsim: The original goal of this doc is to provide instructions to help pre 2.7 users rotate both their trust root and issuer certificate. Do we need a separate 2.7 upgrade note and a generic how-to for issuer cert rotation?

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
ihcsim: Good catch! My bad for providing @zaharidichev with incorrect info.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
ihcsim: We are thinking more of the application data plane, which we have less control over. We feel it's better to tell the users that this process might potentially incur downtime, and they end up with none, than the other way around i.e. we tell them there won't be downtime, but this process brings down their services due to factors we didn't account for.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
ihcsim: It must be used with the `--no-password` flag.

ihcsim:
@adleong From what I gathered, the workflow outlined in your doc is consistent with that in this PR. In addition, your doc provides some additional detailed checks and feedback. My recommendation is that we merge #3696 after all the pending items are addressed, and then use your doc to create a second set of issues to address your recommendations. WDYT?

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
adleong: We could make this more general to encompass both: "Rotating the identity certificates"

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
++++
+title = "Rotating the identity trust root"
+description = "Follow this work flow to update your identity trust root"
```
adleong: "your identity certificates"

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
++++
+
+By default, the trust root that Linkerd uses has a default validity period of
```
adleong: "the issuer certificate and trust root that Linkerd uses are valid for 365 days."

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+365 days. This certificate is stored in the `linkerd-config` config map. If
```
adleong: I think we could use another sentence explaining the context:

"If either of these certificates expires, Linkerd will no longer be able to proxy traffic.  Therefore, it is critical that you replace these certificates with new ones before they expire -- a process called certificate rotation."

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+your Linkerd control plane was installed using either the `linkerd install`,
+`linkerd install --ha` or `helm install` command, this documentation provides
+information to help you to rotate the trust root certificate before it expires.
+
+If your control plane is installed with the `linkerd install
+--identity-external-issuer` command where your trust root is managed by a 3rd
+party certificate management solution like `cert-manager`, then this information
+doesn't apply to you.
```
adleong: "... because it is the responsibility of your certificate manager to rotate the certificates before they expire."

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
adleong: I just don't think there's any reason to sow doubt.  We don't advertise that Linkerd can cause downtime in general, and this process is no different.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+Running this command at the end of each step and ensuring that all checks pass
+should give you sufficient guarantee that you are progressing in the right
+direction.
+
+{{< note >}}
+There is likelihood that this process may incur downtime. Therefore it is
+advisable to plan for that happening.
+{{< /note >}}
+
+## Generate new trust root and issuer certificates
```
adleong: Before getting to this step, we should explain that if you installed with a manually supplied trust root, and still have the root key, and your trust root isn't expiring, then you can simply use your root key to generate a new issuer and skip directly to "Updating the identity issuer certificate".

Otherwise, you'll need to generate a new trust root, add that trust root to the cluster, and use that newly generated root key to create a new issuer.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+
+First generate the root certificate with its private key. The private key needs
+to be stored in a secured vault so it can be used to generate a new trust root
+in the future.
+
+```bash
+step certificate create identity.linkerd.cluster.local ca-new.crt ca-new.key --profile root-ca --no-password --insecure
```
adleong: `--insecure` looks like such a scary flag, we should perhaps provide some explanation of what this flag does and reassurance that this isn't actually insecure.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+Note that the above command will generate a root certificate. Alternately, you
```
adleong: I think we can skip this "Alternately" sentence if we talk about it above, as I suggested.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+can use your existing trust root to generate an intermediate certificate for
+Linkerd to use as its mTLS trust root.
+
+Now generate the intermediate certificate and key pair:
```
adleong: I think it flows better if we move this step down into the "Updating the identity issuer certificate" section

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+step certificate create identity.linkerd.cluster.local issuer-new.crt issuer-new.key --ca ca-new.crt --ca-key ca-new.key --profile intermediate-ca --not-after 8760h --no-password --insecure
+```
+
+## Bundling your original trust root with the new one
+
+As a next step we need to bundle the trust root currently used by Linkerd
+together with the new root and configure Linkerd to use the new bundle:
+
+```bash
+kubectl -n linkerd get cm linkerd-config -o=jsonpath='{.data.global}' |  \
```
adleong: this command looks intimidating, perhaps we can just add a comment that this fetches the current trust root from the linkerd configmap.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+jq -r .identityContext.trustAnchorsPem > original-trust.crt &&
+step certificate bundle ca-new.crt original-trust.crt bundle.crt &&
```
adleong: and this step bundles together the existing trust root with the new one to form the `bundle.crt` bundle file

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+
+```bash
+kubectl -n emojivoto rollout restart deploy
+```
+
+Now you can run the `check` command to ensure that everything is ok:
+
+```bash
+linkerd check --proxy
+```
```
adleong: perhaps include some sample output of what it looks like while still waiting for the pods to restart

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+
+You might have to wait a few moments until all the pods have been restarted and
+are configured with the correct roots. When that becomes true, you should stop
+seeing TLS related warning being output by the `check` command.
```
adleong: perhaps include sample output of what it looks like once the trust rollout is complete

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Updating the identity issuer certificate
+
+Now that all proxies are updated to include the additional trust root, it is
```
adleong: As I mentioned above, I think we can move the step where we generate the issuer into this section here.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+can check for the specific Kubernetes event.
+
+```bash
+kubectl get events --field-selector reason=IssuerUpdated -n linkerd
+
+LAST SEEN   TYPE     REASON          OBJECT                        MESSAGE
+9s          Normal   IssuerUpdated   deployment/linkerd-identity   Updated identity issuer
+```
+
+Run `linkerd check --proxy` to make sure that everything is going as expected.
```
adleong: Provide sample output and note that the warnings about certs expiring soon are gone.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Removing the old trust root
+
+At this point, it is safe to remove the old bundle and restart the proxies once
```
adleong: Perhaps we can note here that the cert rotation is complete and all warnings should be gone.  But for hygiene, we can clean up the old trust root.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
adleong: Many folks will not read any of the words in this guide, and just go down the page copy-pasting each command.  So I think explicitly including a check command after each step is a good idea.

linkerd.io/content/2/tasks/rotating_identity_trust_root.md
```diff
@@ -0,0 +1,166 @@
...
+again. Using the `upgrade` command can do that for the Linkerd components:
+
+```bash
+bin/linkerd upgrade  --identity-trust-anchors-file=./ca-new.crt  | kubectl apply -f -
```
adleong: remember to use the `linkerd` command instead of `bin/linkerd` in all the examples

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+{{< note >}}
+Although Linkerd can auto-generate the trust root during installation, we
+recommend using your own trust root for all serious workloads, so that you
+have full control over it. See [Generating your own mTLS root certificates](/2/tasks/generate-certificates/#generating-the-certificates-with-step)
+on how to do this.
+{{< /note >}}
+
+## Prerequisites
+
+We are going to use `step` (0.13.3) and `jq` (1.6). We will also make use of the
```
adleong: should we include a link to where one can download these tools?

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+`linkerd check` CLI utility which will give us extra assurance that the state of
```
adleong: We should call out that the `linkerd` CLI must be version at least stable-2.7.0.  Is there a minimum control plane version?  Just the release where mTLS was introduced, I guess?

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+our mTLS configuration is valid at any point. To begin with, you can run:
+
+```bash
+linkerd check --proxy
+```
+
+You should see output similar to:
```
adleong: If your configuration is valid and your certificates are not expiring soon, you should see output similar to:
...

However, if you see the message "X" it means that you must perform certificate rotation before your certificates expire.

If you see "Y", it means that your certificates have already expired.  See Z (separate section in this doc, or perhaps a separate doc) for how to recover from this state.

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+```bash
+linkerd-identity
+----------------
+√ certificate config is valid
+√ trust roots are using supported crypto algorithm
+√ trust roots are within their validity period
+√ trust roots are valid for at least 60 days
+√ issuer cert is using supported crypto algorithm
+√ issuer cert is within its validity period
+√ issuer cert is valid for at least 60 days
```
adleong: perhaps these sample outputs should include warnings about the issuer expiring soon so that we can see these warnings go away when we update the issuer.

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+√ issuer cert is issued by the trust root
+
+linkerd-identity-data-plane
+---------------------------
+√ data plane proxies certificate match CA
+```
```
adleong: I think we need a separate section in this doc or an entirely separate doc on the workflow for when you have a certificate that is already expired.  If it's okay with you, can you create that as a separate PR so that it doesn't slow this PR down from getting merged?

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,239 @@
...
+
+## Removing the old trust root
+
+At this point the cert rotation process is complete. However as you do not need
```
ihcsim: I think we should emphasize on the needs to remove the old trust root from the bundle. Can we say something like:

> At this point, the cert rotation process is complete. For security purposes, we will remove the old trust root from the trust bundle we created earlier. The `linkerd upgrade` command can do that.
> ```bash
> linkerd upgrade  --identity-trust-anchors-file=./ca-new.crt  | kubectl apply -f -
> ```
> Note that the `./ca-new.crt` file is the same trust root you created at the start of this process.

zaharidichev:
@adleong I added the page here as these two link to each other and I think they genrally go hand in had. Feel free to share your thoughts on it as well.

zaharidichev:
I am not sure I understand why these checks fail since the website links work locally: 

```
========================================================================
running in concurrent mode, this is experimental
2/tasks/replacing_expired_certificates/index.html
  hash does not exist --- 2/tasks/replacing_expired_certificates/index.html --> /2/tasks/rotating_identity_certificates/#updating-the-identity-issuer-certificate
2/tasks/rotating_identity_certificates/index.html
  hash does not exist --- 2/tasks/rotating_identity_certificates/index.html --> /2/tasks/rotating_identity_certificates/#updating-the-identity-issuer-certificate
1/advanced/namerd/index.html
  Non-OK status: 404 --- 1/advanced/namerd/index.html --> https://twitter.github.io/commons/apidocs/com/twitter/common/zookeeper/ServerSet.html
```

linkerd.io/content/2/tasks/replacing_expired_certificates.md
```diff
@@ -0,0 +1,137 @@
...
+    Invalid roots:
+        * 272080721524060688352608293567629376512 identity.linkerd.cluster.local not valid anymore. Expired on 2019-12-19T10:05:31Z
+    see https://linkerd.io/checks/#l5d-identity-roots-are-time-valid for hints
+```
+
+You can follow [Generating your own mTLS root certificates](/2/tasks/generate-certificates/#generating-the-certificates-with-step)
+to create new root and issuer certificates. Then use the `linkerd upgrade`
+command:
+
+```bash
```
adleong: I think it's a bit confusing that we give this command without the `--force` flag at first.  I understand that you included it to educate on why the `--force` flag is necessary, but I think we can just explain that with words.  As it is, this almost makes it seem like running without `--force` at first is a necessary prerequisite step.

linkerd.io/content/2/tasks/rotating_identity_certificates.md
```diff
@@ -0,0 +1,297 @@
...
+As a next step we need to bundle the trust root currently used by Linkerd
+together with the new root and configure Linkerd to use the bundle. The
+following command uses `kubectl` to fetch the Linkerd config as a `json`
+object, extracts the current roots from the config with the help of  `jq` and
+finally uses `step` to combine it with the newly generated roots and save the
+result in the `bundle.crt` file.
+
+```bash
+kubectl -n linkerd get cm linkerd-config -o=jsonpath='{.data.global}' |  \
+jq -r .identityContext.trustAnchorsPem > original-trust.crt &&
```
adleong: do we need to chain these commands together with `&&` or can we just run them sequentially?

Issue #619, opened 2020-01-02
clarify 2.6 release notes data plane upgrade requirement


wmorgan:
The 2.6 release notes say: "Once your control plane is successfully upgraded, your data plane must be restarted."

AIUI, there is no requirement that your data plane *must* be restarted, and as written it sounds like you need to immediately take an uptime hit while you reboot everything. I suggest removing this sentence or clarifying where it lies in spectrum from "it would be nice" to "mandatory".

PR #621, opened 2020-01-04
Clarify data-plane restart requirements

cpretzer:

@grampelberg, per our conversation I ran these upgrade tests:
- 2.4.0 to 2.6.0
- 2.5.0 to 2.6.0

In both cases, the 2.4.0 and 2.5.0 proxies running under the 2.6.0 control plane continued to run without error.

I also confirmed that a pod injected with a 2.4.0 proxy could successfully communicate with a pod injected with a 2.6.0 proxy.

I'll run the same tests with 2.6.1, and for now, the tests support the documentation.

### DIFF
--- a/linkerd.io/content/2/tasks/upgrade.md
+++ b/linkerd.io/content/2/tasks/upgrade.md
@@ -95,7 +95,10 @@ to ensure zero downtime upgrade for previous stable releases.
 
 The `destination` container is now deployed as its own `Deployment` workload.
 Once your control plane is successfully upgraded, your data plane must be
-restarted.
+restarted if you are upgrading from one of the edge versions listed above.
+
+If you are upgrading from a previous stable version, restarting the data-plane
+is __recommended__ as a best practice, although not necessary.
 
 If you have previously labelled any of your namespaces with the
 `linkerd.io/is-control-plane` label so that their pod creation events are
--- a/linkerd.io/content/2/tasks/upgrade.md
+++ b/linkerd.io/content/2/tasks/upgrade.md
@@ -94,8 +94,11 @@ to ensure zero downtime upgrade for previous stable releases.
 {{< /note >}}
 
 The `destination` container is now deployed as its own `Deployment` workload.
-Once your control plane is successfully upgraded, your data plane must be
-restarted if you are upgrading from one of the edge versions listed above.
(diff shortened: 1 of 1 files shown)


cpretzer:
fixes #619

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -34,7 +34,10 @@ to ensure zero downtime upgrade for previous stable releases.
 
 The `destination` container is now deployed as its own `Deployment` workload.
 Once your control plane is successfully upgraded, your data plane must be
-restarted.
+restarted if you are upgrading from one of the edge versions listed above.
+
+If you are upgrading from a previous stable version, restarting the data-plane
+is __recommended__ as a best practice, although not necessary.
```
ihcsim: But **eventually**, the data plane will need to be restarted in order to pull down the latest version of the proxy, right? The _although not necessary_ made the restart sounds completely optional. My understanding is restart is needed, just not immediately.

Since there is nothing special about the 2.6 upgrade restart process, will it work if we just remove the entire sentence about restarting i.e. just ends with _the destination controller is now deployed..._?

cpretzer: @ihcsim you're right and I think we should point out that the proxies will not be updated until they are restarted.

Rather than omit the sentence about restarting, I like the idea of being explicit about the consequences of _not_ restarting.

WDYT?

ihcsim: Those info are provided in the [Upgrade the data plane](https://linkerd.io/2/tasks/upgrade/#upgrade-the-data-plane ) section. The Upgrade Notice provides release-specific highlights (breaking changes etc.) to the users. It then walks them through how to upgrade the CLI, the control plane and data plane, which remain the same across releases. 

If you feel there is a need to be more explicit for 2.6, I'm fine with it.

cpretzer: 👍 thanks @ihcsim

grampelberg:
@cpretzer I'm still pretty uncomfortable with how you've worded this. You *must* upgrade the data plane to the same version as the control plane, you can just take your time. WDYT about leaving the must terminology and simply saying you can do it over the course of a day or week?

cpretzer:
I'm good with that. I'll make the change

PR #643, opened 2020-01-22
Update Helm install doc with renamed value

alpeb:
globals.noInitContainer is now globals.cniEnabled

This is to be merged after linkerd/linkerd2#3956 merges

### DIFF
--- a/linkerd.io/content/2/tasks/install-helm.md
+++ b/linkerd.io/content/2/tasks/install-helm.md
@@ -53,10 +53,8 @@ file passed with a `-f` option, or overriding specific values using the family o
 
 ## Disabling The Proxy Init Container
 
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/tasks/install-helm.md
```diff
@@ -53,10 +53,8 @@ file passed with a `-f` option, or overriding specific values using the family o
 
 ## Disabling The Proxy Init Container
 
-If installing with CNI, make sure that you add the
-`--set global.noInitContainer=true` and
-`--set installNamespace=false` flags to your `helm install` command as these
```
zaharidichev: Nice catch that this flag is not required anymore as the cni plugin is in a different namespace by default

PR #647, opened 2020-01-27
Elaborate on helm upgrade

zaharidichev:
Turns out we cannot simply upgrade the helm chart in case of breaking changes to it. There are bit more details that ned to be considered. Therefore I expanded on the docs a bit to keep the users informed. Currently ` helm upgrade linkerd2 linkerd/linkerd2 --reuse-values` when moving from 2.6 to edge does not work. 

Signed-off-by: Zahari Dichev <zaharidichev@gmail.com>

### DIFF
--- a/linkerd.io/content/2/tasks/install-helm.md
+++ b/linkerd.io/content/2/tasks/install-helm.md
@@ -116,8 +116,36 @@ NAME                    CHART VERSION          APP VERSION            DESCRIPTIO
 linkerd/linkerd2        <chart-semver-version> {{% latestversion %}}    Linkerd gives you observability, reliability, and securit...
 ```
 
-Use the `helm upgrade` command to upgrade the chart:
+The `helm upgrade` command has a number of flags that allow you to customize
+its behaviour. The ones that special attention should be paid to are
+`--reuse-values` and `--reset-values` and how they behave when charts change
+from version to version and/or overrides are applied through `--set` and
+`--set-file`. To summarize there are the following prominent cases that can be
+observed:
+
+- `--reuse-values` with no overrides - all values are reused
+- `--reuse-values` with overrides - all except the values that are overridden
+are reused
+- `--reset-values` with no overrides - no values are reused and all changes
+from provided release are applied during the upgrade
+- `--reset-values` with overrides - no values are reused and changed from
+provided release are applied together with the overrides
+- no flag and no overrides - `--reuse-values` will be used by default
+- no flag and overrides - `--reset-values` will be used by default
+
+Bearing all that in mind, you have to decide whether you want to reuse the
+values in the chart or move to the values specified in the newer chart.
+The advised practice is to use a `values.yaml` file that stores all custom
+overrides that you have for your chart. Before upgrade, check whether there
+are breaking changes to the chart (i.e. renamed or moved keys, etc). You can
+consult the [edge](https://hub.helm.sh/charts/linkerd2-edge/linkerd2) or the
+[stable](https://hub.helm.sh/charts/linkerd2/linkerd2) chart docs, depending on
+which one your are upgrading to. If there are, make the corresponding changes to
+your `values.yaml` file. Then you can use:
 
 ```bash
-helm upgrade linkerd2 linkerd/linkerd2 --reuse-values
+helm upgrade linkerd2 linkerd/linkerd2 --reset-values -f values.yaml --atomic
 ```
+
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2/tasks/install-helm.md
```diff
@@ -116,8 +116,36 @@ NAME                    CHART VERSION          APP VERSION            DESCRIPTIO
 linkerd/linkerd2        <chart-semver-version> {{% latestversion %}}    Linkerd gives you observability, reliability, and securit...
 ```
 
-Use the `helm upgrade` command to upgrade the chart:
+The `helm upgrade` command has a number of flags that allow you to customize
+its behaviour. The ones that special attention should be paid to are
+`--reuse-values` and `--reset-values` and how they behave when charts change
+from version to version and/or overrides are applied through `--set` and
+`--set-file`. To summarize there are the following prominent cases that can be
```
alpeb: Perhaps better `these are the following...`?

linkerd.io/content/2/tasks/install-helm.md
```diff
@@ -116,8 +116,36 @@ NAME                    CHART VERSION          APP VERSION            DESCRIPTIO
...
+overrides that you have for your chart. Before upgrade, check whether there
+are breaking changes to the chart (i.e. renamed or moved keys, etc). You can
+consult the [edge](https://hub.helm.sh/charts/linkerd2-edge/linkerd2) or the
+[stable](https://hub.helm.sh/charts/linkerd2/linkerd2) chart docs, depending on
+which one your are upgrading to. If there are, make the corresponding changes to
+your `values.yaml` file. Then you can use:
 
 ```bash
-helm upgrade linkerd2 linkerd/linkerd2 --reuse-values
+helm upgrade linkerd2 linkerd/linkerd2 --reset-values -f values.yaml --atomic
```
alpeb: If I understood correctly, `--reset-values` here is not needed?

zaharidichev:
@alpeb 

> On the other hand, this leads me to think whether we should list all the chart values changes in the release notes...

I actually think this is the best idea... Maybe not for edges but for stables, it would make things easier. Nevertheless this is not a light process for the user at all... Maybe I am missing something here and there is an easier way to do all that

linkerd.io/content/2/tasks/install-helm.md
zaharidichev: It is not, I jut put it for completeness, but I can remove it

alpeb:
For this particular upgrade (2.6 -> 2.7) we might also provide some extra tooling to ease the transition. Turns out one can use `jq` as a [template engine](https://github.com/stedolan/jq/wiki/Cookbook#using-jq-as-a-template-engine), so we could use the output from `helm get values --output json` and transform it into the new format. I'd be glad to give that a shot.

alpeb:
... More concretely, to transform the values from 2.6 to 2.7 one would do:
```
helm get values -a --output json release-name | jq -f values_2.7_map.json
```
Using this [values_2.7_map.json](https://gist.github.com/alpeb/346c651e08f5a6954cb801b638b17abf)

zaharidichev:
@alpeb That is quite neat. I wonder what our approach should be here. Maybe just include a section here examplaining how to perform such transformation and then include the json map file with the release?

zaharidichev:
@alpeb So per our conversation from a few days ago I think what we need to do about that is the following: 
1. In this section here, avoid mentioning anything about --reset and --reuse and simply explain that the best way to upgrade is to keep a custom values.yaml around that specifies all modiciation to the chart
2. Explaing that this values need to be provided on every helm --upgrade
3. Note that if there are breaking changes that happen in the chart we will publish a diff file in the upgrade notes, which can be used to figure out has moved/changed (link to the upgrade files)

alpeb:
@zaharidichev I thought the section on flags and overrides was wonderful, as I haven't found that elsewhere; not even the official docs are clear about it!
But yeah, this might not be the right place to delve into Helm specifics (but I'll have to keep a copy for myself :stuck_out_tongue: ). Keeping explanations as succinct as possible sounds good to me :+1:

zaharidichev:
@alpeb Yes its a pitty it has to go as it was not easy to understand all the details. But yes, I will ammend the PR to be a bit more laconic

wmorgan:
Can we merge this now since 2.7 is a Helm-breaking release?

grampelberg:
I'm gonna merge, if we have any other feedback it can be taken care of in a follow up PR.

PR #659, opened 2020-02-10
fix canonical og urls

wmorgan:
Per https://ogp.me/, we should be setting `og:url` to

> The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "http://www.imdb.com/title/tt0117500/".

Currently we set it to `https://linkerd.io` for everything, which breaks sharing on LinkedIn.

This patch changes it to what I *believe* is the actual canonical URL for the page.


wmorgan:
Ooops, lemme fix this

PR #668, opened 2020-02-24
Remove hard-coded channel versions

kleimkuhler:
## Motivation

As part of the release automation project, removing hard-coded versions from
`linkerd` repositories has been helpful in order to automate certain steps.

The website publish step currently relies on hard-coded channel version
values.

## Solution

`curl` the releases of `linkerd2` and pull out the first matching `edge-..`
and `stable-..` tags. Use those values in the `make get-versions` target which
is used by other targets such as `make publish`.

This allows for a website related workflow to automatically update the install
script only by triggering a `make publish` job.

Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>


Makefile
```diff
@@ -21,12 +24,16 @@ HAS_HTMLTEST := $(shell command -v htmltest;)
...
-publish: update-version build-linkerd.io deploy
+publish: get-versions build-linkerd.io deploy
 	@# Publish a new version of the sites
 
-.PHONY: update-version
-update-version: replace-env-L5D2_STABLE_VERSION replace-env-L5D2_EDGE_VERSION
+.PHONY: get-versions
+get-versions:
 	@# Update the version for the %* site
+	L5D2_STABLE_VERSION=$(shell ./bin/channel-version stable)
```
grampelberg: I think you want:

```bash
$(MAKE) replace-env-L5D2_STABLE_VERSION replace-env-L5D2_EDGE_VERSION \
    L5D2_STABLE_VERSION=$$(./bin/channel-version stable) \
    L5D2_EDGE_VERSION=$$(./bin/channel-version edge)
```

kleimkuhler: Okay the script is called once now and just exports the version values into the environment.

Issue #671, opened 2020-02-26
Add contour ingress documentation


arminbuerkle:
When https://github.com/linkerd/linkerd2/pull/4092 is merged Linkerd will support contour/envoy as ingress controller. 

To accommodate this change the https://linkerd.io/2/tasks/using-ingress/ website needs to be updated to explain to users how to use contour in combination with linkerd.

PR #672, opened 2020-02-26
Add contour ingress documentation

arminbuerkle:
Refs #671

Documentation for https://github.com/linkerd/linkerd2/pull/4092


stefanprodan:
@grampelberg can we add a note to the docs? If Contour is used together with Flagger, the `l5d-dst-override` is added [automatically](https://github.com/weaveworks/flagger/blob/master/pkg/router/contour.go#L48)

grampelberg:
@stefanprodan that's awesome, totally should have something like that in there.

arminbuerkle:
Rebased and added a note about flagger. AFAIK the failing tests are not a result of this PR.

PR #684, opened 2020-03-09
use a single period to avoid the enforced host check

lundbird:
Signed-off-by: alex lundberg <alex.lundberg@commonbond.co>

### DIFF
--- a/linkerd.io/content/2/tasks/exposing-dashboard.md
+++ b/linkerd.io/content/2/tasks/exposing-dashboard.md
@@ -133,7 +133,7 @@ spec:
         - -enforced-host=^dashboard\.example\.com$
 ```
 
-If you want to completely disable the `Host` header check, use a single period
+If you want to completely disable the `Host` header check, use one period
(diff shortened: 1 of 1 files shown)


lundbird:
Can we rerun the validate check? It doesnt seem to be related.

lundbird:
I added another commit and it passes linting

alpeb:
@lundbird can you please explain the reasoning behind this change?

lundbird:
You explain here  https://github.com/linkerd/linkerd2/pull/4154 that the default value for enforced-host is an empty string which allows request if you port-forward. This is also the default in the values.yaml. This prevents access for us when we access the dashboard through a custom domain name.

We were able to get around this by using a single period for the regex match instead of an empty string.

lundbird:
Nevermind I see what the confusion is about. I can add some more description to clear what this check does with an empty string and what "completely disable" means.

cpretzer:
Thanks for the explanation @lundbird I'll keep an eye out for the changes

PR #685, opened 2020-03-09
update helm install docs to use linkerd/ instead of charts/ as the base

lundbird:
Signed-off-by: alex lundberg <alex.lundberg@commonbond.co>


alpeb:
These are the cert generation docs, so just linking against the[ full helm installation docs ](https://linkerd.io/2/tasks/install-helm/)for more info should do the trick.

grampelberg:
I'm +1 to that!

PR #689, opened 2020-03-16
Describe how to achieve longer lived trust anchor certificates.

naseemkullah:
It is in some cases desirable to have a longer than default lived trust anchor certificate for long lived clusters.

Also added headers to distinguish between Trust anchor creation and Issuer Cert/Key creation.

### DIFF
--- a/linkerd.io/content/2/tasks/automatically-rotating-control-plane-tls-credentials.md
+++ b/linkerd.io/content/2/tasks/automatically-rotating-control-plane-tls-credentials.md
@@ -55,6 +55,9 @@ step certificate create identity.linkerd.cluster.local ca.crt ca.key \
    --namespace=linkerd
 ```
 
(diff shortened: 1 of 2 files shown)


grampelberg:
@naseemkullah it appears that github can't check the code out?

naseemkullah:
@grampelberg Seems like a case of https://github.com/actions/checkout/issues/23

I will try re running a build forcibly by pushing en empty commit.

naseemkullah:
Error same as original: https://github.com/linkerd/website/runs/514853817?check_suite_focus=true

Re-running build will surely error with that described in actions/checkout#23

cpretzer:
Thanks for sticking with this @naseemkullah 

The current failures are related to an aggressive check on links in unrelated docs, so I'm going to push this through

naseemkullah:
My pleasure @cpretzer !

PR #700, opened 2020-03-27
Add server speaks first ports

naseemkullah:
Cassandra internode communication and Redis Sentinel.

There is no clear documentation as to these protocol's behaviour in terms of server speaks first, but they had to be ignored to keep things working.

### DIFF
--- a/linkerd.io/content/2/features/protocol-detection.md
+++ b/linkerd.io/content/2/features/protocol-detection.md
@@ -52,16 +52,16 @@ For example, if your application makes requests to a MySQL database running on
 port 4406, use the command:
 
 ```bash
-linkerd inject deployment.yml --skip-outbound-ports=4406 \
-  | kubectl apply -f -
+linkerd inject deployment.yml --skip-outbound-ports=4406 |
+  kubectl apply -f -
 ```
 
 Likewise if your application runs an SMTP server that accepts incoming requests
 on port 35, use the command:
 
 ```bash
-linkerd inject deployment.yml --skip-inbound-ports=35 \
-  | kubectl apply -f -
+linkerd inject deployment.yml --skip-inbound-ports=35 |
+  kubectl apply -f -
 ```
 
 ## Plaintext MySQL and SMTP
--- a/linkerd.io/content/2/features/protocol-detection.md
+++ b/linkerd.io/content/2/features/protocol-detection.md
@@ -38,6 +38,12 @@ The following protocols are known to be server-speaks-first:
 * 4222 - NATS
 * 27017 - MongoDB
 
+The following protocols are likely but to be confirmed to be
+server-speaks-first(users have had to ignore these ports):
+
+* 7000 - Cassandra inter-node cluster communication
+* 26379 - Redis Sentinel
+
 If you're working with a protocol that can't be automatically recognized by
 Linkerd, use the `--skip-inbound-ports` and `--skip-outbound-ports` flags when
 running `linkerd inject`.
--- a/linkerd.io/content/2/features/protocol-detection.md
+++ b/linkerd.io/content/2/features/protocol-detection.md
@@ -38,8 +38,8 @@ The following protocols are known to be server-speaks-first:
 * 4222 - NATS
 * 27017 - MongoDB
 
-The following protocols are likely but to be confirmed to be
-server-speaks-first(users have had to ignore these ports):
+The following protocols are most likely (but yet to be confirmed)
+server-speaks-first:
 
 * 7000 - Cassandra inter-node cluster communication
 * 26379 - Redis Sentinel


linkerd.io/content/2/features/protocol-detection.md
```diff
@@ -46,16 +52,16 @@ For example, if your application makes requests to a MySQL database running on
 port 4406, use the command:
 
 ```bash
-linkerd inject deployment.yml --skip-outbound-ports=4406 \
-  | kubectl apply -f -
+linkerd inject deployment.yml --skip-outbound-ports=4406 |
+  kubectl apply -f -
```
cpretzer: This is meant to be on two lines in order to show that they are two separate commands.

linkerd.io/content/2/features/protocol-detection.md
```diff
@@ -46,16 +52,16 @@ For example, if your application makes requests to a MySQL database running on
...
 ```
 
 Likewise if your application runs an SMTP server that accepts incoming requests
 on port 35, use the command:
 
 ```bash
-linkerd inject deployment.yml --skip-inbound-ports=35 \
-  | kubectl apply -f -
+linkerd inject deployment.yml --skip-inbound-ports=35 |
+  kubectl apply -f -
```
cpretzer: Same as the last comment. Let's keep these as two separate lines

linkerd.io/content/2/features/protocol-detection.md
```diff
@@ -38,6 +38,12 @@ The following protocols are known to be server-speaks-first:
 * 4222 - NATS
 * 27017 - MongoDB
 
+The following protocols are most likely (but yet to be confirmed)
+server-speaks-first:
```
cpretzer: Thanks for taking the time to put this together ⭐️ 

This could say `Some examples of server-speaks-first protocols are:`

I'd prefer to add only protocols that are _known_ to be server speaks first

linkerd.io/content/2/features/protocol-detection.md
naseemkullah: Do you agree that the proposed change is technically correct?
The `\` is for stylistic reasons?

linkerd.io/content/2/features/protocol-detection.md
naseemkullah: I cannot confirm that due to lack of documentation on those protocols, but they dont work with linkerd 🤷‍♂  is that enough to lump them in with "known to be" ?

linkerd.io/content/2/features/protocol-detection.md
cpretzer: I'm not sure that the proposed change will work if someone copies the text and pastes it into their terminal. I suspect that the line break will carry over from the HTML and the shell will not handle it properly.

I'd have to try it for myself :)

naseemkullah: but `|` works the same way `\` does to continue a command across the next line

linkerd.io/content/2/features/protocol-detection.md
cpretzer: I'm struggling with this one because the docs should be accurate. I'd prefer not to have protocols mentioned based on suspicion of something being server speaks first.

That being said, specifying that those _might_ be server speaks first could help with troubleshooting. Perhaps this would fit better in a "Checking for server speaks first protocols" in the Troubleshooting docs?

We could link to that doc from here. 

WDYT?

naseemkullah: Do you mean to add a "Checking for server speaks first protocols" in the Troubleshooting docs? I probably would not be qualified to write that section as I do not know how to check unless it's clear in the protocol's docs.

BTW if you are or know a C wiz who could confirm Sentinel code seems to be here: https://github.com/antirez/redis/blob/29b9d0a245135a0b9e024032a7cccd2c238103a9/src/sentinel.c

linkerd.io/content/2/features/protocol-detection.md
cpretzer: Oh, cool. TIL! I've always used a slash.

Stylistically, the slash is used throughout the docs. If we change it here, then we should change it everywhere for consistency and I don't think it's worth the change since they produce the same result.

linkerd.io/content/2/features/protocol-detection.md
cpretzer: Yeah, I was suggesting a new section in the troubleshooting docs to help identify server speaks first issues.

@grampelberg WDYT?

I'm not a C whiz unfortunately, however, I also found some redis docs which describe a connection turning in to a server push protocol when Pub/Sub is used:

```When a Redis client subscribes to a Pub/Sub channel, the protocol changes semantics and becomes a push protocol, that is, the client no longer requires sending commands, because the server will automatically send to the client new messages (for the channels the client is subscribed to) as soon as they are received.```

I suspect that would be a port that we want to skip.

linkerd.io/content/2/features/protocol-detection.md
grampelberg: My personal preference is to leave it as is. Feels superfluous for this specific update anyways =)

naseemkullah: I also just very recently discovered this via my IDE's shell script formatter @cpretzer !
I guess there is an argument for the added `\` in terms of clarity. Will leave it as is! Infact as per other comments will close this PR until the server-speaks-first section in troubleshooting is added.

linkerd.io/content/2/features/protocol-detection.md
naseemkullah: sgtm will close this for now until server-speaks-first can be confirmed for these ports. Thanks!

PR #726, opened 2020-05-14
Updates for the multicluster tutorial

grampelberg:


grampelberg:
@zaharidichev I ran out of time today, I'll jump onto this tomorrow. Can you make sure what I've got in here makes sense and the flow works? I've got a couple `TODO` comments around commands that I'd love help with if you've got them.

grampelberg:
I also need to update all the images, they're kinda placeholders right now I stole from the blog posts. I've got the draw.io templates for those though, so it'll be easy for me to update.

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,282 @@
...
+    alt="step-1"
+    title="Copy Services"
+    center="true"
+    src="/images/multicluster/step-1.svg" >}}
+
+At this point, you'll have the mirror installed on your `london` cluster. It
+won't be able to do anything yet, we'll get to that in a minute. Make sure that
+everything is up and running with:
+
+TODO: better command here.
```
zaharidichev, line 83: `maybe linkerd check --multicluster`. That should verify the controller is up an running, has all the needed RBAcs, etc

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,282 @@
...
+{{< fig
+    alt="step-3"
+    title="Gateway"
+    center="true"
+    src="/images/multicluster/step-3.svg" >}}
+
+You can see what's happened by running:
+
+```bash
+TODO: command to verify
```
zaharidichev, line 124: you are trying to verify that the gateway is installed and everything works, right? Not sure what the best command for that would be. Do we need separate checks for the seprate clusters- local cluster specific checks, remote cluster specific checks ?

grampelberg:
@zaharidichev in my defense it is a literary reference to a tale of two cities =) Do you think that makes it a little bit more clear? I got super wrapped up trying to differentiate between clusters named local/remote and the concepts of local/remote.

grampelberg:
@zaharidichev is there a way to verify that the mirror has picked up the secret and is actively watching? I ran `check`, but it won't fail if there were issues reading the secret AFAIK.

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,451 @@
...
+`london` and `paris` clusters now.
+
+```bash
+linkerd install \
+  --identity-trust-anchors-file root.crt \
+  --identity-issuer-certificate-file issuer.crt \
+  --identity-issuer-key-file issuer.key \
+  | tee \
+    >(kubectl --context=london apply -f -) \
+    >(kubectl --context=paris apply -f -)
```
olix0r, line 111: It'd be better to use different issuer credentials in each cluster, in my opinion. but i understand that might be a distraction.

grampelberg: I guess I could get a couple for loops in here like I'm doing further down. Either way, I think we need to educate here about why you shouldn't do it. Given the note around education, is it worth the added complexity? I'm a teensy worried folks will think this setup stuff is all magic if we go too far.

olix0r: I think it's a great opportunity to address that _these_ are the credentials that should be rotated -- this is like the most common question we're getting right now.

cpretzer:
I added some changes to the `base/frontend.yml` because GKE has a restrictive file system that OOTB nginx won't start on.

The changes force files to write to the /tmp directory which is mapped to an `emptyDir` and also set the `runAsUser:1000`

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,730 @@
...
+Installed into the `linkerd-multicluster` namespace, the gateway is a simple
+[NGINX proxy](https://github.com/linkerd/linkerd2/blob/master/charts/linkerd2-multicluster-remote-setup/templates/gateway.yaml#L10)
+which has been injected with the Linkerd proxy. On the inbound side, Linkerd
+takes care of validating that the connection uses a TLS certificate that is part
+of the trust anchor. NGINX takes the request and forwards it to the Linkerd
+proxy's outbound side. At this point, the Linkerd proxy is operating like any
+other in the data plane and forwards the requests to the correct service. Make
+sure the gateway comes up successfully by running:
+
+TODO: not sure this will work in the laptop -> cloud case. TODO: should this all
```
zaharidichev, line 180: It will not work with a vanilla kind installation unless there is the capability get Loadbalancer service. We can either add that to the requirements or simply say that they need to try it with two cloud environemnts. Or point them to a script that creates a k3d setup for them (similar to what @olix0r has)

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,581 @@
...
+on a set of target clusters and copying these services from the target cluster
+to the local one running the service mirror. Instead of creating a new, special
+way to address and interact with services, Linkerd leverages Kubernetes services
+so that your application code does not need to change and there is nothing new
+to learn.
+
+In addition, there needs to be a component that can be reached external to the
+cluster. Requests on one cluster can be routed through the gateway component to
+any service inside another cluster. This gateway will be exposed to the public
+internet via. a `Service` of type `LoadBalancer`. Only requests verified through
```
alpeb, line 148: typo: `via.`

linkerd.io/content/2/tasks/multicluster.md
```diff
@@ -0,0 +1,581 @@
...
+There are two components required to discover services and route traffic between
+clusters. The service mirror is responsible for continuously monitoring services
+on a set of target clusters and copying these services from the target cluster
+to the local one running the service mirror. Instead of creating a new, special
+way to address and interact with services, Linkerd leverages Kubernetes services
+so that your application code does not need to change and there is nothing new
+to learn.
+
+In addition, there needs to be a component that can be reached external to the
+cluster. Requests on one cluster can be routed through the gateway component to
```
alpeb, line 146: ...that can be reached from outside the cluster?

PR #732, opened 2020-05-29
update install commands to be Hem v3

### LINKED ISSUE #730: Helm install commands fails with Helm3
Currently, Most helm commands in the helm tutorial use `--name`, which seems depreciated in Helm3 and the install command fails as 
```
Error: unknown flag: --name
```

In Helm3, The install command works as `helm install [NAME] [CHART]`. So, removing the `--name` flag but still keeping the value fixes this.

Should we update the docs to Helm3? or Should we atleast have a note for the same?

@alpeb

Pothulapati:
Fixes #730 

This PR makes the default docs to represent Helm v3 commands. It also adds a additional note to install commands for users trying with Helm v2.

The main problem being the depreciation of `--name` flag in Helm v3.

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>


grampelberg:
@Pothulapati mind updating this so we can merge?

PR #736, opened 2020-06-08
Fix port for web-svc

ewolff:
The Kubernetes service `web-svc` listens to port 80. So this definition needs to be changed. It is already port 80 in the other Ingress definitions. Also I checked the "Getting Started" emojivote Kubernetes deployment.


ewolff:
@grampelberg Thanks for the nice feedback - I just fixed DCO , so all should be fine now 🙂

ewolff:
Not sure why the validate check fails, though.

grampelberg:
@ewolff that's my fault. Look's great!

PR #737, opened 2020-06-08
Adding a doc on installing multicluster

grampelberg:


linkerd.io/content/2/tasks/installing-multicluster.md
```diff
@@ -0,0 +1,292 @@
...
+mirror.linkerd.io/gateway-ns: linkerd-multicluster
+```
+
+{{< /note >}}
+
+## Leverage Ambassador
+
+The bundled Linkerd gateway is not required. In fact, if you have an existing
+Ambassador installation, it is easy to use it instead! This guide assumes that
+Ambassador has been installed into the `ambassador` namespace.
```
richarddli, line 107: How about adding one more sentence:

By using your existing Ambassador installation, you avoid needing to manage two different ingress gateways.

linkerd.io/content/2/tasks/installing-multicluster.md
```diff
@@ -0,0 +1,292 @@
...
+
+First, you'll want to inject the ambassador deployment with Linkerd:
```
richarddli, line 109: Capitalize "Ambassador"

PR #750, opened 2020-06-18
added note about @grpc/grpc-js bug (linkerd2 #4605)

seeruk:
Original issue reported at https://github.com/linkerd/linkerd2/issues/4605

Signed-off-by: Elliot Wright <wright.elliot@gmail.com>

(PS: I think the HTML validation error is unrelated)


cpretzer:
@seeruk You're right, the html failure is unrelated to this change. I've made some changes that should allow the tests to pass once all the changes are made

PR #751, opened 2020-06-23
Update nginx ingress docs regarding default backends

adleong:
As verified in https://github.com/linkerd/linkerd2/issues/4636


grampelberg:
@adleong mind fixing up lint and then we can get this merged?

PR #752, opened 2020-06-24
proxy-configuration page description clipping fix

tbrockman:
Image depicting issue (Chrome):
linkerd_before

Image depicting layout after fix (Chrome):
linkerd_after

Allow annotation to wrap, preventing description from clipping on longer annotations. Also makes the page more readable at half-width.

Signed-off-by: Theodore Brockman <tbrockma@ualberta.ca>

### DIFF
--- a/linkerd.io/layouts/partials/cli/annotations.html
+++ b/linkerd.io/layouts/partials/cli/annotations.html
@@ -1,6 +1,6 @@
 <tr>
   <td>
-    <code class="text-nowrap">{{ .Name }}</code>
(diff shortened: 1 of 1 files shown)


tbrockman:
No problem @cpretzer !

Yeah, that was my assumption as well, and it's possible this is just a result of longer annotations being added over time, but in any case this seemed like a fine compromise to make for the sake of readability. An alternate layout could also be attempted to accommodate both (maybe something using just horizontal rows) but the table presentation seems pretty natural to me.

PR #754, opened 2020-06-26
clean up upgrade notes

wmorgan:
- add a stable-2.8.0 section so that it doesn't look like we forgot
- move version notes to the end, since they are getting very long


linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -9,9 +9,11 @@ aliases = [
 
 In this guide, we'll walk you through how to upgrade Linkerd.
 
-The upgrade notices below contain important information you need to be aware of
-before commencing with the upgrade process:
+Before starting, read through the version-specific upgrade notices below, whihc
```
grampelberg, line 12: which

linkerd.io/content/2/tasks/upgrade.md
```diff
@@ -25,6 +27,168 @@ There are three components that need to be upgraded:
...
+linkerd check --proxy
+```
+
+This will run through a set of checks against both your control plane and data
+plane to verify that it is operating correctly.
+
+You can make sure that you've fully upgraded all the data plane by running:
+
+```bash
+kubectl get po --all-namespaces -o yaml \
```
grampelberg, line 169: Didn't we do something with check to make this irrelevant? @alpeb ?

alpeb: Right, if there are proxies on an older version you'd see something like this:
```bash
$ linkerd check --proxy
...
linkerd-data-plane
------------------
√ data plane namespace exists
√ data plane proxies are ready
√ data plane proxy metrics are present in Prometheus
‼ data plane is up-to-date
    Some data plane pods are not running the current version:
        * emojivoto/vote-bot-5f887b4c76-mlsf4 (stable-2.8.0)
        * emojivoto/web-8d985c8fb-8q9jr (stable-2.8.0)
        * emojivoto/emoji-7bd4d5655-f55nq (stable-2.8.0)
        * emojivoto/voting-8c7445b4-sjhz8 (stable-2.8.0)
    see https://linkerd.io/checks/#l5d-data-plane-version for hints
```
(when the control plane is on stable-2.8.1)

wmorgan: I just replaced that section with a note about how check --proxy (which we were telling them to run anyways) will report older proxy versions.

PR #770, opened 2020-07-17
Update GKE firewall setup instructions

samstride:
Subject
Updating instructions to create firewall rules for GKE.

Problem
Cannot setup Linkerd on a GKE private cluster due to a variable name change in GCP.

Solution
Update docs to use the current variable. The old variable was masterIpv4CidrBlock. The new variable is clusterIpv4CidrBlock

Signed-off-by: Vijay <vijaysavanth@gmail.com>

### DIFF
--- a/linkerd.io/content/2/reference/cluster-configuration.md
+++ b/linkerd.io/content/2/reference/cluster-configuration.md
@@ -22,11 +22,11 @@ CLUSTER_NAME=your-cluster-name
 gcloud config set compute/zone your-zone-or-region
 ```
 
-Get the cluster `MASTER_IPV4_CIDR`:
+Get the cluster `CLUSTER_IPV4_CIDR`:
 
 ```bash
-MASTER_IPV4_CIDR=$(gcloud container clusters describe $CLUSTER_NAME \
-  | grep "masterIpv4CidrBlock: " \
+CLUSTER_IPV4_CIDR=$(gcloud container clusters describe $CLUSTER_NAME \
+  | grep "clusterIpv4CidrBlock: " \
   | awk '{print $2}')
 ```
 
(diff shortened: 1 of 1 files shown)


cpretzer:
@samstride regarding the HTML validation test failure, I cloned your repository and got the `make lint` command to pass by adding a newline character to `linkerd.io/content/_index.md`

Your repository looks up to date, but would you mind rebasing from master as a sanity check?

samstride:
@cpretzer , ahhh my bad, I missed the difference between VPC-native cluster vs private cluster. I think it's best to create separate docs for VPC-native cluster.

How should we organize it though? Should I add it to the [current page](https://linkerd.io/2/reference/cluster-configuration/) under a new section or create a new page?

ihcsim:
Do we actually need to tweak the firewall rules for VPC native GKE? I just created one with no issues. The doc in this PR pertains only to private GKE clusters where by default, the k8s api server can't communicate with the admission webhooks that get scheduled on the worker nodes.

samstride:
@ihcsim , when I run `linkerd check --pre`, 2 warnings were issued on my GKE cluster around `NET_ADMIN` and `NET_RAW` capability. What is the best way to resolve those warnings or is it safe to ignore them?

Stept to reproduce.

- Create a new GKE cluster with default settings.
- Setup kubectl
- Run `linkerd check --pre`

I assumed that `NET_ADMIN` and `NET_RAW` are required to be able to launch dashboard `linkerd dashboard &` from my local machine. Is that assumption wrong?

ihcsim:
The warnings mean the default pod security policy created by GKE don't have to `NET_ADMIN` and `NET_RAW` capabilities, which are required by the Linkerd `proxy-init` init container, or CNI plugin. See https://linkerd.io/2/tasks/using-psp/. If the VPC native GKE cluster works for you, without the need to change the firewall rules, we should close this PR. LMK.

samstride:
@ihcsim . Thanks for clarifying. Closing PR and issue.

PR #782, opened 2020-08-04
Update multicluster docs for 2.9

adleong:
Do not merge until Linkerd 2.9 is released.

Open question: is this an appropriate place for the multicluster update docs for 2.8 -> 2.9?


linkerd.io/content/2/tasks/upgrade-multicluster.md
```diff
@@ -0,0 +1,107 @@
...
+that mirror services remain up during the upgrade so as to avoid downtime.
+
+### Control Plane
+
+Begin by upgrading the Linkerd control plane itself to version 2.9 by running
+
+```bash
+linkerd upgrade | kubectl apply -f -
+```
+
```
alpeb, line 51: Don't the source and target clusters also require a `linkerd mc install` as well?

alpeb (commented):
Thanks @adleong :+1: 

The [/features/multicluster doc](https://github.com/linkerd/website/blob/master/linkerd.io/content/2/features/multicluster.md) also requires a tiny change to use labels instead of annotations :wink:

linkerd.io/content/2/tasks/installing-multicluster.md
```diff
@@ -197,20 +193,17 @@ metadata:
...
+With everything setup and configured, you're ready to link a source cluster to
+this Ambassador gateway.  Run the `link` command specifying the name and
+namespace of your Ambassador service:
 
-```yaml
-mirror.linkerd.io/gateway-name: ambassador
-mirror.linkerd.io/gateway-ns: ambassador
+```bash
+linkerd --context=${ctx} multicluster link --gateway-name=ambassador --gateway-namespace=ambassador \
+    | kubectl --context=${src_ctx} apply -f -
```
alpeb, line 202: `linkerd multicluster link` now requires always specifying `--cluster-name`

alpeb:
The annotation for the Ambassador service is lacking the leading slash, which avoids the service mirror to properly hit it:
https://github.com/linkerd/website/pull/782/files#diff-fae783921db4c9259ab03260024736ceb39c4f77feca6a5c46d79e1d7b98930fL195

The error seen in the service mirror controller looks like this:
```
level=error msg="Could not create a GET request to gateway: parse \"http://probe-gateway-east:80-/ambassador/ready\": invalid port \":80-\" after host" probe-key=east
```

alpeb:
> To fetch your existing cluster's trust anchor, run:
> ```
> kubectl -n linkerd get cm/linkerd-config -ojsonpath="{.data.global}" | \
>  jq .identityContext.trustAnchorsPem -r > trustAnchor.crt
> ```

This is no longer applicable in 2.9 after the config changes. I can only think of using `yq` here instead:
```
kubectl -n linkerd get cm linkerd-config -ojsonpath="{.data.values}" | yq -r .global.identityTrustAnchorsPEM > trustAnchor.crt
```

linkerd.io/content/2/tasks/installing-multicluster.md
```diff
@@ -337,21 +327,20 @@ The installation can be verified by running
...
+`linkerd multicluster install` command through the CLI.
 
-The same functionality can also be done through Helm by disabling `gateway` and
-`serviceMirror` while submitting the remote service account name.
+The same functionality can also be done through Helm by disabling `gateway`
+while submitting the remote service account name.
 
 ```bash
- helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set gateway=false --set serviceMirror=false --set remoteMirrorServiceAccountName=source --set installNamespace=false --kube-context target
+ helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set gateway=false --set remoteMirrorServiceAccountName=source --set installNamespace=false --kube-context target
```
alpeb, line 343: This command is generating an error. I created #5098 explaining the situation.

alpeb:
In `multicluster.md` under the "Cleanup" section we're still instructing running `kubectl --context=${ctx} delete ns linkerd-multicluster`, but now we have `linkerd --context=${ctx} multicluster uninstall | kubectl --context=${ctx} delete -f -` which will also remove the cluster-level resources.

linkerd.io/content/2/tasks/upgrade-multicluster.md
adleong: I think this might not be necessary because nothing actually changed between 2.8 and 2.9 except for the link and service mirror machinery.  We also need to be careful not to remove the old service mirror until the new one is in place.  But we should walk through this process to be sure.

linkerd.io/content/2/tasks/installing-multicluster.md
```diff
@@ -337,34 +336,26 @@ The installation can be verified by running
...
+through the CLI.
 
-The same functionality can also be done through Helm by disabling `gateway` and
-`serviceMirror` while submitting the remote service account name.
+The same functionality can also be done through Helm setting the
+`remoteMirrorServiceAccountName` value to a list.
 
 ```bash
- helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set gateway=false --set serviceMirror=false --set remoteMirrorServiceAccountName=source --set installNamespace=false --kube-context target
+ helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set remoteMirrorServiceAccountName={source1,source2,source3} --kube-context target
```
alpeb, line 356: From my testing, the commas require escaping:
```
helm install linkerd2-mc-source linkerd/linkerd2-multicluster --set remoteMirrorServiceAccountName={source1\,source2\,source3} --kube-context target
```

adleong: Yeah, though I think this might be shell dependent.  I was following the style of the Helm docs which don't escape them: https://github.com/helm/helm/blob/7cbed3f77bd771dfb46c63733c497e8c05225422/docs/using_helm.md#the-format-and-limitations-of---set

linkerd.io/content/2/tasks/upgrade-multicluster.md
alpeb: Turns out we need the link CRD definition in the source cluster for `linkerd mc link` below to work.Thankfully, `linkerd mc install` doesn't kill the old service mirror (but upgrading through helm would).
And to avoid having `linkerd check --multicluster` throwing an error on the target (also because the CRD def is missing), we could also upgrade mc there.

alpeb (approved):
Thanks @adleong !

PR #785, opened 2020-08-05
Support ARM CLI

aliariff:
# Changes
- Add checking the machine arch for an arm machine and download the correct binary file.

Merge before https://github.com/linkerd/linkerd2/pull/4841

### DIFF
--- a/run.linkerd.io/public/install
+++ b/run.linkerd.io/public/install
@@ -118,6 +118,7 @@ fi
 
   echo "Downloading ${srcfile}..."
   if ! curl -fLO "${url}"; then
+    # shellcheck disable=SC2030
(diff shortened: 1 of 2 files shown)


grampelberg:
Would you mind adding tests for these as well?

aliariff:
The broken test is expected. The current release did not have an extra arch name in the release artifacts.
After #4841 merged and a new release tag is created, then this will pass.

aliariff:
And adding a test for the arm machine, I don't think we can do that, because the script relies on `uname -m` which we cant stub.

kleimkuhler:
@aliariff Just a heads up it looks like this needs to merge with the latest changes on master

PR #786, opened 2020-08-06
fixing external prometheus setup documentation

mirkop-mattr:
* API endpoint for prometheus is generally under / and not under /api/prom ( this endpoint is for writing back to prometheus)
* when using external prometheus the sample frequency needs to be higher than for example what is the default in prometheus helm chart 
* Using the already deployed prometheus configuration is a good starting point as a reference for an external prometheus setup.

### DIFF
--- a/linkerd.io/content/2/tasks/external-prometheus.md
+++ b/linkerd.io/content/2/tasks/external-prometheus.md
@@ -124,7 +124,8 @@ The running configuration of the builtin prometheus can be used as a reference.
 kubectl -n linkerd  get configmap linkerd-prometheus-config -o yaml
 ```
 
-In order for the dashboard and grafana to work correctly the scrape settings need to be set to the following values: 
+In order for the dashboard and grafana to work correctly the scrape settings need
+to be set to the following values:
 
 ```yaml
 scrape_interval: 10s
--- a/linkerd.io/content/2/tasks/external-prometheus.md
+++ b/linkerd.io/content/2/tasks/external-prometheus.md
@@ -36,8 +36,11 @@ Before applying, it is important to replace templated values (present in `{{}}`)
 with direct values for the below configuration to work.
 
 ```yaml
-
     - job_name: 'linkerd-controller'
+
+      scrape_interval: 10s
+      scrape_timeout: 10s
+
       kubernetes_sd_configs:
       - role: pod
         namespaces:
@@ -53,6 +56,10 @@ with direct values for the below configuration to work.
         target_label: component
 
     - job_name: 'linkerd-service-mirror'
+
+      scrape_interval: 10s
+      scrape_timeout: 10s
+
       kubernetes_sd_configs:
       - role: pod
       relabel_configs:
@@ -66,6 +73,10 @@ with direct values for the below configuration to work.
         target_label: component
 
     - job_name: 'linkerd-proxy'
+
+      scrape_interval: 10s
+      scrape_timeout: 10s
+
       kubernetes_sd_configs:
       - role: pod
       relabel_configs:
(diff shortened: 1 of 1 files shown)


mirkop-mattr:
FYI The HELM instructions are not valid for the current helm chart in the linkerd helm chart repo. They only work when using the latest chart from github.

Pothulapati:
@mirkop-mattr Users can get the latest **edge** chart from https://linkerd.io/2/tasks/install-helm/#adding-linkerds-helm-repository by adding the edge helm repo instead of the normal one. Does that not work?

linkerd.io/content/2/tasks/external-prometheus.md
```diff
@@ -118,6 +118,22 @@ with direct values for the below configuration to work.
...
+
+```bash
+kubectl -n linkerd  get configmap linkerd-prometheus-config -o yaml
+```
+
+In order for the dashboard and grafana to work correctly the scrape settings need
+to be set to the following values:
+
+```yaml
+scrape_interval: 10s
```
Pothulapati, line 131: Hmm How about we combine this configuration with the above same configuration, and provide a single one? @alpeb

linkerd.io/content/2/tasks/external-prometheus.md
```diff
@@ -134,14 +150,9 @@ which is available both through `linkerd install` and `linkerd upgrade` commands
 
 ```yaml
 global:
-  prometheusUrl: existing-prometheus.xyz:9090/api/prom
+  prometheusUrl: existing-prometheus.xyz:9090
```
Pothulapati, line 153: This is correct! :+1: I mistook the URL with the other remote-write thing that I was working on but I remember the dashboard working. so, definitely going to give this another spin!

linkerd.io/content/2/tasks/external-prometheus.md
mirkop-mattr: Yep sure thing. I have updated the PR.

linkerd.io/content/2/tasks/external-prometheus.md
mirkop-mattr: > @mirkop-mattr Users can get the latest **edge** chart from https://linkerd.io/2/tasks/install-helm/#adding-linkerds-helm-repository by adding the edge helm repo instead of the normal one. Does that not work?

Yes, its availabe in edge. But the entire documentation references the stable version. I assume with 2.9 coming out soon this will not be an issue anymore.

mirkop-mattr: @Pothulapati Can you please help with the failed HTML validator? The issue is with a file that i have not touched.

ihcsim:
@mirkop-mattr Thanks for the fix! Mind taking a look at the [DCO](https://github.com/linkerd/website/blob/master/CONTRIBUTING.md#developer-certificate-of-origin) and [amend the commit](https://github.com/linkerd/website/pull/786/checks?check_run_id=955038487) so that CI passes? 

PS The other CI failures aren't related to this PR - they are just sporadic timeout errors caused by endpoint checks in older blog posts.

mirkop-mattr:
@ihcsim Apologies I should have read the  doc first. I will update the commit.

mirkop-mattr:
@Pothulapati I was wondering if the prometheus code snippet should just be changed so it works out of the box rather than having the template on the docs page.

ihcsim:
@mirkop-mattr I think having the config template in the doc is fine. I prefer users can just follow the doc from start to finish, to achieve their goal, without having to navigate to GitHub, or run `kubectl get` or `helm pull` etc.

PR #788, opened 2020-08-07
figo merged into finleap

jsoref:
Fix `Lint html and check for dead links` as reported in https://github.com/linkerd/website/runs/955191928


cpretzer:
Hi @jsoref, thanks for this change.

Would you mind amending the commit to include your digital signature? It's important that we get that and you can learn about the process [here](https://github.com/linkerd/linkerd2/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #797, opened 2020-08-13
Document cert-manager experimental controller issue

MaT1g3R:
Running linkerd versions <= stable-2.8.1 with cert-manager version 0.15
experimental controllers enabled causes linkerd-identity pods to crash
with this error message:

> "Failed to initialize identity service: failed to read CA from disk: unsupported block type: 'PRIVATE KEY'"

This commit documents some potential solutions to this problem.

Signed-off-by: Peijun Ma <peijun.ma@protonmail.com>


ihcsim:
@MaT1g3R Mind fixing up the linter error? Thanks.

PR #798, opened 2020-08-14
Add data plan proxy check in multicluster setup step

secret104278:
resolve https://github.com/linkerd/linkerd2/issues/4875

Signed-off-by: Yu Chih <secret104278@gmail.com>

### DIFF
--- a/linkerd.io/content/2/tasks/installing-multicluster.md
+++ b/linkerd.io/content/2/tasks/installing-multicluster.md
@@ -49,6 +49,7 @@ To verify that everything has started up successfully, run:
 
 ```bash
 linkerd check --multicluster
+linkerd -n linkerd-multicluster check --proxy
 ```
 
 For a deep dive into what components are being added to your cluster and how all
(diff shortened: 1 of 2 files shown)


ihcsim:
@secret104278 Thanks for the PR. I think a better place for this change is the [troubleshooting guide](https://linkerd.io/2/tasks/troubleshooting/), since the `linkerd check` command already provides a link to the relevant section:

```
× all target cluster gateways are alive
    Some gateways are not alive:
    * cluster: [east], gateway: [linkerd-multicluster/linkerd-gateway]
    see https://linkerd.io/checks/#l5d-multicluster-target-gateways-alive for hints
```

We can add a new paragraph that says something like "use `linkerd check --proxy -n linkerd-multicluster` to make sure the gateway proxies are healthy."

(The correct link is actually https://linkerd.io/2/tasks/troubleshooting/#l5d-multicluster-remote-gateways-alive.)

LMKWYT.

PR #808, opened 2020-09-09
More SEO fixes

PatrickHeneise:

PR #811, opened 2020-09-15
Update README to include Hugo version

ihcsim:


ihcsim:
I'm gonna go ahead and merge this PR, since only the README is changed. It's a follow-up of #808 where the Hugo version is bumped.

PR #813, opened 2020-09-17
Linkerd with Argo CD GitOps doc

### LINKED ISSUE #757: Gitops with Linkerd
Deploying and managing Linkerd in a fully gitops fashion can be a little confusing. Let's get a example up in the [repo](https://github.com/linkerd/linkerd-examples) that runs through the full walkthrough as part of a Makefile or set of scripts. The end goal of this is to write some documentation for the website, for now it should be an example repo that someone can run through themselves.

# Demo

1. Run git server on cluster
    - Show pulling and pushing from repo
1. Install flux
    - Show flux running
1. Install [helm-operator](https://github.com/fluxcd/helm-operator-get-started)
    - Show helm operator running.
1. Configure flux to use on-cluster git server
    - Show flux syncing on-cluster git server
1. Create linkerd certificates
    - Show certificate details
1. Setup cert-manager
    - Validate cert-manager install
1. Add linkerd certificates to cluster via cert-manager (is there a way to manage cert-manager via gitops?).
    - Show cert-manager created certificates, inspect to see what they look like.
1. Add linkerd configuration to repo (via helm operator).
    - Show control plane deploying, validate with `check`.
1. Add emojivoto to repo.
    - Show emojivoto metrics via linkerd dashboard.

## Extra credit

1. Upgrade linkerd control-plane
    - Validate new version has been installed
1. Upgrade proxy version in the data-plane
    - Validate new version has been installed.

# Requirements

- It should be possible to cut/paste each step (multiple cut/pastes are fine). Users shouldn't need to go figure out how to set something up, instead they should have it setup on their cluster for them. This is the reasoning behind putting a git server on the user's cluster for them.
- Every step needs validation. It should be clear to the user that they have successfully completed the step and provide a way for them to understand exactly what that step did for them.

ihcsim:
This PR adds a guide with instructions on how to set up Argo CD, cert-manager and SealedSecrets to manage the installation and upgrade of Linkerd. The steps in this guide will work only after https://github.com/linkerd/linkerd-examples/pull/261 is merged.

Fix #757.


linkerd.io/content/2/tasks/gitops.md
```diff
@@ -0,0 +1,534 @@
...
+diff -b \
+  <(echo "${trust_anchor}" | step certificate inspect -) \
+  <(step certificate inspect sample-trust.crt)
+```
+
+### Deploy Linkerd
+
+Now we are ready to install Linkerd. The decrypted trust anchor we just
+retrieved will be passed to the installation process using the
+`global.identityTrustAnchorsPEM` parameter.
```
muenchdo, line 369: I think it would be great to explain why we need to explicitly pass the trust anchor here again. My initial thought when reading this was "Wait, isn't this certificate already stored in the `linkerd-trust-anchor` that we deployed using sealed secrets?".

muenchdo:
Not sure if this goes beyond the scope of this guide, but it would be nice to know what the steps would be to deploy Linkerd to a second cluster. Since the trust anchor needs to be shared for multicluster, could the final output of this just be deployed to more clusters and we'd be good to go?

linkerd.io/content/2/tasks/gitops.md
ihcsim: Good point on adding an explanation for passing the trust anchor again. This guide follows the flow defined in [automatic cert rotation](https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/) and [generating root certificate](https://linkerd.io/2/tasks/generate-certificates/) docs. Instead of duplicating content here, what do you think of adding a link in this section to those docs, and if needed, add more explanation there?

As for setting this parameter from the start, which sub-section of the doc do you think will be more appropriate? The `linkerd` Argo application is deployed (but not sync-ed) as part of the `main` parent application, before we use `step` to generate the root certificate. `step` is first used in the `linkerd-bootstrap` application sub-section. My hope is that by not mingle the instructions for the `linkerd` application with that of the `linkerd-bootstrap` application, it will be clearer to the users which application is being worked on in each sub-section.

muenchdo: Personally, I tend to dislike guides which refer to other guides, it's usually never a smooth experience going through them. It could be enough to just explain that we put the trust anchor into a secret so that cert-manager can use it to create the issuer certificates, but that Linkerd itself does not (cannot?) read the trust anchor from the secret, so it needs to be "statically injected" as well.

How about starting out with generating the root certificate ("First we need to generate a root certificate which we will use in two places: For enabling cert-manager to act as a CA for creating/rotating issuer certificates, and for ...") and then doing the rest?

ihcsim: > but that Linkerd itself does not (cannot?) read the trust anchor from the secret, so it needs to be "statically injected" as well.

Yup, that sounds like a good addition. I usually use links to refer to additional, detailed information that I deem isn't required to complete the task at-hand.

> How about starting out with generating the root certificate

I am not convinced this is better. SealedSecret will need to be deployed and sync-ed before we can perform the YAML and git operations of `linkerd-bootstrap`. That will require the `main` application to be deployed beforehand, which would have deployed (not not sync-ed) `linkerd-bootstrap` and `linkerd` anyway.

grampelberg (approved):

Issue #826, opened 2020-10-16
Automatic mTLS issuer cert rotation only works for k8s 1.16.14+ 


cpackingham:
I was trying to follow the automatic mTLS rotation instructions on the website with our 1.15.11 cluster, and ran into issues where the examples for cert-manager didn't run on the specified api versions. I downgraded to v1 on the resources, but ran linkerd check and the check failed here:
```
× issuer cert is using supported crypto algorithm
    issuer certificate must use ECDSA for public key algorithm, instead RSA was used
    see https://linkerd.io/checks/#l5d-identity-issuer-cert-uses-supported-crypto for hints
```

Replication steps:

##1.15.11
```
minikube start --profile=v1.15.11 --kubernetes-version=v1.15.11

kubectl create namespace linkerd || echo "linkerd namespace already exists"

step certificate create identity.linkerd.cluster.local ca.crt ca.key \
--profile root-ca --no-password --insecure --san identity.linkerd.cluster.local &&
  kubectl create secret tls \
   linkerd-trust-anchor \
   --cert=ca.crt \
   --key=ca.key \
   --namespace=linkerd


## cert-manager
kubectl create namespace cert-manager || echo "cert-manager namespace already exists"
helm repo add jetstack https://charts.jetstack.io
helm repo update
helm install \
  cert-manager jetstack/cert-manager \
  --namespace cert-manager \
  --version v1.0.3 \
  --set installCRDs=true


## had to downgrade api version to v1 for this, v1alpha2 and v1alpha3 didn't exist on 1.15.11
kubectl apply -f linkerd/common/cert-issuer.yaml
kubectl apply -f linkerd/common/cert.yaml

## linkerd
helm repo add linkerd https://helm.linkerd.io/stable

helm install linkerd2 \
--set-file global.identityTrustAnchorsPEM=ca.crt \
--set identity.issuer.scheme=kubernetes.io/tls \
--set installNamespace=false \ 
-n linkerd \
linkerd/linkerd2

linkerd check
```
Should fail checks for issuer certs

##1.16.14
```
minikube start --profile=v1.16.14 --kubernetes-version=v1.16.14

kubectl create namespace linkerd || echo "linkerd namespace already exists"

step certificate create identity.linkerd.cluster.local ca.crt ca.key \
--profile root-ca --no-password --insecure --san identity.linkerd.cluster.local &&
  kubectl create secret tls \
   linkerd-trust-anchor \
   --cert=ca.crt \
   --key=ca.key \
[...]

PR #827, opened 2020-10-19
Fix line of bash and add version compatibility note for cert-manager

### LINKED ISSUE #825: Fix shell line in automatic certificate rotation task
The line of bash at the bottom of [automatically rotating control plane tls credentials](https://linkerd.io/2/tasks/automatically-rotating-control-plane-tls-credentials/#installing-with-helm) is not runnable - the name of the helm chart should be on the same line as `helm install`. Pretty minor thing but also easy to fix!

cpackingham:
Fixes #825, #826


grampelberg:
Perfect, ty!

PR #828, opened 2020-10-21
update ingress apiVersion

WLun001:
`extensions/v1beta1` is deprecated


WLun001:
Did not notice the API breaking change on v1

PR #835, opened 2020-10-27
add note for externally rendered helm charts

cypherfox:
Clusters that use helm to externally render templates seem to be more susceptible to a race condition when rotating certificates in for the webhooks. This MR adds minimal notes to warn about this.

Signed-off-by: Lutz Behnke <lutz.behnke@finleap.com>


adleong (approved):

alpeb (commented):
@cypherfox From my testing, `helm upgrade` will always update the secrets and all the webhooks will be restarted to pick up the changes (since the introduction of the `checksum/config` annotation). So no need to delete the secrets under any scenario using Helm.
Also, would you rename the secrets as you did for the cert-manager compatibility PR that you pushed?
Finally, it'd be nice if at the beginning of this doc it was pointed out that there's an automated way of doing this, pointing to the doc created by #837

alpeb:
Actually I added a note to #837 wondering if we should replace this doc with the one from that PR... so let's see first if this doc gets deprecated before making more changes :pray:

cypherfox:
> @cypherfox From my testing, `helm upgrade` will always update the secrets and all the webhooks will be restarted to pick up the changes (since the introduction of the `checksum/config` annotation). So no need to delete the secrets under any scenario using Helm.

I cannot prove it yet, but I have some indications that this success relies on the order the resources are defined in the actual l5d helm charts. Thus it will work with straight helm, but not always with helm-rendering and then applying the resources (e.g. with ArgoCD or our homegrown solution) 

> Also, would you rename the secrets as you did for the cert-manager compatibility PR that you pushed?

Will do

> Finally, it'd be nice if at the beginning of this doc it was pointed out that there's an automated way of doing this, pointing to the doc created by #837

Will do as well.

alpeb:
@cypherfox we finally decided to deprecate this doc in favor of the one introduced in #837 . Do you mind looking into that one and see if some of your comments are applicable there?

grampelberg:
Also, see #852.

PR #840, opened 2020-10-28
Update Ingress Docs about new ingress mode

Pothulapati:
This PR adds a new section into the using ingress doc, about the new
ingress mode along with explanation on what the current two ways in
which a ingress controller can run with Linkerd.

Signed-off-by: Tarun Pothulapati <tarunpothulapati@outlook.com>


alpeb (commented):
The additional explanations look solid :+1: 
For more clarity, I think it might make sense to divide the whole doc into two sections, e.g. "Default Mode" and "Proxy Ingress Mode". The latter would start at "If you're planning on injecting Linkerd into your ingress controller's pods...". The current subsections (Nginx, Traefik, etc.) would need to become subsections of "Proxy Ingress Mode".

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,10 +3,34 @@ title = "Using Ingress"
 description = "Linkerd works alongside your ingress controller of choice."
 +++
 
+When your ingress controller is injected, The load balancing decision would
+be taken by the Ingress Controller and the proxy just routes it to the
```
adleong, line 7: perhaps clarify "Linkerd proxy" instead of just "proxy"

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,10 +3,34 @@ title = "Using Ingress"
...
+choosen target Pod IP by the Ingress controller. This means that Linkerd functionality
+like Service Profiles, Traffic Splits would not work as the current discovery is
+based on the Service IP. TCP mTLS and Load Balancing should still work however.
```
adleong, line 10: s/should/will

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,10 +3,34 @@ title = "Using Ingress"
...
+
+```bash
+kubectl get deployment <ingress-controller> -n <ingress-namespace> -o yaml | linkerd inject --ingress - | kubectl apply -f -
+```
+
+This can be verified by checking if the `linkerd-proxy` has the relevant environment
+variable set by the `proxy-injector`.
+
+```bash
+kubectl describe pod/<ingress-pod> | grep LINKERD2_PROXY_INGRESS_MODE
```
adleong, line 32: can we look at the annotation in the pod spec of the deployment instead? it would be nice to avoid referencing proxy env variables in public facing documentation

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
-If you're planning on injecting Linkerd into your ingress controller's pods
-there is some configuration required. Linkerd discovers services based on the
+From Linkerd `v2.9`, There are two ways in which Linkerd proxy
+can be ran with your Ingress Controller.
+
+## Default Mode
+
+When the ingress controller is injected, The load balancing decision would
+be taken by the Ingress Controller and the Linkerd proxy just routes it to the
+choosen target Pod IP by the Ingress controller. This means that Linkerd functionality
```
kleimkuhler, line 13: This first sentence felt a little wordy, but not a blocker:

```suggestion
When the ingress controller is injected, it makes the load balancing decision by choosing the target pod IP and the Linkerd proxy routes it. This means that Linkerd functionality
```

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
+like Service Profiles, Traffic Splits would not work as the current discovery is
```
kleimkuhler, line 14:
```suggestion
like Service Profiles, Traffic Splits would not work as its discovery is
```

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
+based on the Service IP. TCP mTLS and Load Balancing will still work however.
```
kleimkuhler, line 15: It may be worth noting here that the load balancing that is being honored is that of the ingress controller and not the proxy's load balancing.

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
+This is useful in cases where you want Linkerd to honour the load balancing
+decision taken by the Ingress Controller for things like Sticky Sessions, etc.
+
+If your Ingress controller is injected with no other configuration specific to
+ingress, The Linkerd proxy runs in the default mode.
```
kleimkuhler, line 20:
```suggestion
ingress, the Linkerd proxy runs in the default mode.
```

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
+
+## Proxy Ingress Mode
+
+If you want Linkerd functionality like Service Profiles, Traffic Splits, etc,
+there is additional configuration required to make the Ingress controller's Linkerd
+proxy run in `ingress` mode, where Linkerd discovers services based on the
 `:authority` or `Host` header. This allows Linkerd to understand what service a
 request is destined for without being dependent on DNS or IPs.
```
kleimkuhler, line 28:
```suggestion
If you want Linkerd functionality like Service Profiles, Traffic Splits, etc, there is additional configuration required to make the Ingress controller’s Linkerd proxy run in ingress mode. This allows Linkerd to discover services based on `:authority` or `Host` headers and understand what service a request is destined for without being dependent on DNS or IPs
```

kleimkuhler (approved):
All my comments are nonblocking hence the approval, but I think we be a little more clear or concise in certain sections! Looks good though 👍

linkerd.io/content/2/tasks/using-ingress.md
Pothulapati: I'm considering just removing `load balancing` here as it's already understood from the previous sentences?

kleimkuhler: Yep that sounds good!

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
-If you're planning on injecting Linkerd into your ingress controller's pods
-there is some configuration required. Linkerd discovers services based on the
-`:authority` or `Host` header. This allows Linkerd to understand what service a
+From Linkerd `v2.9`, There are two ways in which Linkerd proxy
+can be ran with your Ingress Controller.
+
+## Default Mode
+
+When the ingress controller is injected, it makes the balancing
+decision by choosing the target Pod IP and the inkerd proxy just routes it.
```
adleong, line 12: "...is injected with the `linkerd.io/inject: enabled` annotation, the Linkerd proxy will honor load balancing decisions made by the ingress controller instead of applying its own EWMA load balancing.  This also means that the Linkerd proxy will not use Service Profiles for this traffic and therefore will not expose per-route metrics or do traffic splitting."

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,11 +3,47 @@ title = "Using Ingress"
...
+decision taken by the Ingress Controller for things like Sticky Sessions, etc.
+
+If your Ingress controller is injected with no extra configuration specific to
+ingress, the Linkerd proxy runs in the default mode.
+
+## Proxy Ingress Mode
+
+If you want Linkerd functionality like Service Profiles, Traffic Splits, etc,
+there is additional configuration required to make the Ingress controller's Linkerd
+proxy run in `ingress` mode. This allows Linkerd to discover services based on the
```
adleong, line 26: ".. `ingress` mode.  This causes Linkerd to route requests based on their `:authority`, `Host`, or `l5d-dst-override` header instead of their original destination which allows Linkerd to perform its own load balancing and use Service Profiles to expose per-route metrics and enable traffic splitting."

linkerd.io/content/2/tasks/using-ingress.md
```diff
@@ -3,10 +3,45 @@ title = "Using Ingress"
 description = "Linkerd works alongside your ingress controller of choice."
 +++
 
-If you're planning on injecting Linkerd into your ingress controller's pods
-there is some configuration required. Linkerd discovers services based on the
-`:authority` or `Host` header. This allows Linkerd to understand what service a
-request is destined for without being dependent on DNS or IPs.
+As of Linkerd version 2.9, There are two ways in which Linkerd proxy
```
adleong, line 6: "There" should not be capitalized

adleong, line 6: "...in which *the* Linkerd proxy..."

alpeb (approved):

PR #849, opened 2020-11-06
Add release blost

grampelberg:


wmorgan:
Can you please update the banner on the main page as well? Something like "Announcing Linkerd 2.9! mTLS for all TCP, and much more. [Read more &raquo;](link)"

PR #868, opened 2020-12-01
increase identity issuer validity to two days

cypherfox:
When following the current documentation on cert-manager integration for identity issuer rotation (https://linkerd.io/2/tasks/automatically-rotating-webhook-tls-credentials/) may lead to certificates becoming invalid before they can be rotated.

This change will prevent this.

Currently there is already a note to review the validity period, but it does not work out of the box.

The reason for this behaviour not working as documented any more *MAY* be the inclusion of https://github.com/linkerd/linkerd2/pull/3893

Signed-off-by: Lutz Behnke <lutz.behnke@finleap.com>


cpretzer (approved):
Thanks @cypherfox

Issue #879, opened 2020-12-29
Add note to use linkerd2/releases instead of brew linkerd for GODEBUG=x509 error


jsoref:
https://github.com/linkerd/linkerd2/issues/5423#issuecomment-749826305:
```
linkerd check
...
linkerd-webhooks-and-apisvc-tls
-------------------------------
× tap API server has valid cert
    cert is not issued by the trust anchor: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0
    see https://linkerd.io/checks/#l5d-tap-cert-valid for hints
```
The suggestion in https://github.com/linkerd/linkerd2/issues/5423#issuecomment-750463389 is to use https://github.com/linkerd/linkerd2/releases instead of the brew installed version

cpretzer:
This is really helpful, thanks @jsoref

adleong:
Even better would be to update https://linkerd.io/2/tasks/automatically-rotating-webhook-tls-credentials/ to indicate that the dnsNames field should be set so that the SAN field will get populated.  This way, the certificates should work with both go 1.14 and go 1.15 and therefore should work with the Linkerd CLI even if it was installed with brew.

Although... looking at that webpage now, it looks like Certificate resources in the instructions there already have the dnsNames field.  @jsoref does `linkerd check` fail for you when following those instructions?

jsoref:
I got this even w/ a SAN field -- I think we're using those instructions w/ that change...

adleong:
You're getting the Linkerd check error which says that your tap API cert doesn't have a SAN field when your tap API cert *does* have a SAN field?  That's surprising.  Are you able to share your tap API certificate (not the private key)?

adleong:
@jsoref any update on this?

jsoref:
I could share the cert, I'm not sure how helpful this is going to be. We've moved a little forward and are running a version of linkerd where `viz` stuff is split out, and thus the brew version of linkerd that I have here complains about things before it gets to the fun bits, and I'm not really sure which is the corresponding certificate.

I'm updating brew now, but I'm not sure I'll be able to give a useful report. If you have a particular command you'd like me to run, that'd be easier than me trying to guess.

For reference, here's my current status:
```
~/Downloads/linkerd2-cli-edge-21.1.1-darwin check
kubernetes-api
--------------
√ can initialize the client
√ can query the Kubernetes API

kubernetes-version
------------------
[...]

PR #880, opened 2020-12-29
Use linkerd/linkerd2/releases not brew (#879)

jsoref:
The brew version uses a newer stricter version of Go.

Signed-off-by: Josh Soref <jsoref@users.noreply.github.com>


cpretzer (changes requested):
Thanks for this update @jsoref , would you mind updating the URL as a markdown link to fix the check that is failing?

Here's an example from the same doc: https://github.com/linkerd/website/pull/880/files#diff-16d4b387d98d34e5a2f9abd8e0c5bbdb510506796583627d6f9a0af94ba483a7R716

You could even do something like 

```
...downloading the Linkerd CLI from the [releases page](https://github.com/linkerd/linkerd2/releases) on GitHub...
```

jsoref:
@cpretzer: I took your text as written

cpretzer (approved):
LGTM

PR #882, opened 2020-12-30
Update examples to use cert-manager v1

### LINKED ISSUE #881: Update examples to use cert-manager v1
The v1.0 cert-manager release was 4 months ago and I expect many users have already upgraded so the examples should be updated to reflect the v1 Certificate CRD, in particular the keyAlgorithm to privateKey/algorithm change.  There are already lots of helpful notes about incompatibilities in older versions.

erwbgy:
Problem: Examples use older v1alpha3 syntax which has changed slightly in v1

Solution: Update examples to use cert-manager v1 which was released 4 months ago.

Fixes #881 

Signed-off-by: Keith Burdis <keith.burdis@gs.com>


cpretzer:
thanks again @erwbgy I'll run through this next week

cpretzer (approved):
This works well. Thanks @erwbgy

Issue #883, opened 2020-12-30
Set subject of the linkerd-trust-anchor CA certificate to reflect its purpose


erwbgy:
The subject of the linkerd-trust-anchor CA certificate in _Automatically Rotating Control Plane TLS Credentials_ and _Multi-cluster communication_ is currently `identity.linkerd.cluster.local` which is the same as intermediate CA certificate linkerd-identity-issuer and I found this confusing.  The subject should be changed to `linkerd-trust-anchor` to reflect its purpose.

PR #884, opened 2020-12-30
Set subject of the linkerd-trust-anchor CA certificate to reflect its purpose

erwbgy:
Problem: The subject of the linkerd-trust-anchor CA certificate in 'Automatically Rotating Control Plane TLS Credentials' and 'Multi-cluster communication' is currently `identity.linkerd.cluster.local` which is the same as intermediate CA certificate linkerd-identity-issuer and I found this confusing. 

Solution:  The subject should be changed to `linkerd-trust-anchor` to reflect its purpose.

Fixes: #883

Signed-off-by: Keith Burdis <keith.burdis@gs.com>

### DIFF
--- a/linkerd.io/content/2/tasks/automatically-rotating-control-plane-tls-credentials.md
+++ b/linkerd.io/content/2/tasks/automatically-rotating-control-plane-tls-credentials.md
@@ -57,13 +57,13 @@ Next, using the [`step`](https://smallstep.com/cli/) tool, create a signing key
 pair and store it in a Kubernetes Secret in the namespace created above:
 
 ```bash
-step certificate create identity.linkerd.cluster.local ca.crt ca.key \
+step certificate create linkerd-trust-anchor ca.crt ca.key \
   --profile root-ca --no-password --insecure &&
   kubectl create secret tls \
-   linkerd-trust-anchor \
-   --cert=ca.crt \
-   --key=ca.key \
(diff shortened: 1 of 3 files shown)


cpretzer:
thanks @erwbgy anything that makes the docs clearer is helpful.

I'll run through the steps next week to make sure everything works as expected and we can get this merged.

wmorgan (commented):
AIUI the subject of an X.509 cert is intended to be the "thing" that the certificate secures, not the name of the certificate itself. By that definition I'm not sure this change makes sense, since what the cert is securing is still the cluster. Is there another way of clarifying the difference between trust anchor and intermediate certs?

erwbgy:
You're right about the subject, but in the case of a CA certificate the issuer and the subject are the same.  The issuer field identifies the entity that has signed and issued the certificate which in this case is the trust anchor itself.

adleong:
@erwbgy nice catch.  What do think about using something a subject like `root.linkerd.cluster.local` or something along those lines?  This differentiates the root certificate from the issuer certificate but keeps the same naming scheme.

erwbgy:
@adleong Sounds good to me.

erwbgy:
@adleong Thoughts about a SAN matching the Subject being added to all certificates?

cpretzer:
This looks to be superceded by #886 so I'm closing and we can reopen if necessary

PR #898, opened 2021-01-20
Adding zonal and regional config set commands to GKE private cluster configuration settings

magickatt:
Noticed that the `gcloud config set compute/zone` command only works for zonal clusters, so thought might be handy to add the command for regional clusters also since it is slightly different.


adleong (approved):

adleong:
Ah, @magickatt it looks like our linter would like you to limit the line length to 80 characters in this file:

 linkerd.io/content/2/reference/cluster-configuration.md:25:81 MD013/line-length Line length [Expected: 80; Actual: 133]

Would you mind taking care of that?

magickatt:
Sorry it's been a hectic week! For sure I've updated it to use a shorter link

adleong:
Thank you!

alpeb (approved):

PR #918, opened 2021-01-27
Configure htmltest checks to exclude internal anchors

cpretzer:
* Remove the htmltest validation of internal hashes and 
* Add a user-agent header for requests


adleong:
It seems like perhaps https://tools.ietf.org is rate-limiting us.  We'll probably have to add it to the list of excluded urls.

cpretzer:
> It seems like perhaps https://tools.ietf.org is rate-limiting us. We'll probably have to add it to the list of excluded urls.

I'm noticing some long request times (on the order of 10s of seconds) from my home network as well. I'm going to give it a try later and see if the tests will pass on their own. Otherwise, I'll add the domain

cpretzer:
l added https://tools.ietf.org and now the validate workflow isn't running at all. I'll dig into this tomorrow.

adleong (approved):

PR #939, opened 2021-02-23
Added "service mesh" in the first paragraph

CathPag:
@wmorgan we should include a few service mesh references here and there.


kleimkuhler:
Closing in favor of #958

PR #936, opened 2021-02-19
Update edge install script text to account for viz

alpeb:
Fixes linkerd/linkerd2#5773

```console
$ LINKERD2_VERSION=stable-2.9.3 run.linkerd.io/public/install

Validating checksum...
Checksum valid.

Linkerd stable-2.9.3 was already downloaded; making it the default 🎉

To force re-downloading, delete '/home/alpeb/.linkerd2/bin/linkerd-stable-2.9.3' then run me again.

Add the linkerd CLI to your path with:

  export PATH=$PATH:/home/alpeb/.linkerd2/bin

Now run:

  linkerd check --pre                     # validate that Linkerd can be installed
  linkerd install | kubectl apply -f -    # install the control plane into the 'linkerd' namespace
  linkerd check                           # validate everything worked!

You can also obtain observability features by installing the viz extension:

  linkerd viz install | kubectl apply -f -  # install the viz extension into the 'linkerd-viz' namespace
  linkerd viz check                         # validate the extension works!
  linkerd dashboard                         # launch the dashboard

Looking for more? Visit https://linkerd.io/2/next-steps
```

PR #944, opened 2021-02-25
Update stable install script text to account for viz

alpeb:
Followup to #936 for the install script that installs linkerd stable


run.linkerd.io/public/install
```diff
@@ -16,7 +16,12 @@ happyexit() {
   echo "  linkerd check --pre                     # validate that Linkerd can be installed"
   echo "  linkerd install | kubectl apply -f -    # install the control plane into the 'linkerd' namespace"
   echo "  linkerd check                           # validate everything worked!"
-  echo "  linkerd dashboard                       # launch the dashboard"
+  echo ""
+  echo "You can also obtain observability features by installing the viz extension:"
```
wmorgan, line 20: How about "Want the on-cluster metrics pipeline and dashboard? Run:"

wmorgan (approved):
Don't need to bikeshed this but it would be nice if all this output fit into 80 cols without wrapping. I think this could be done easily by trimming some whitespace and removing the namespace from the comments.

alpeb:
@wmorgan I've shortened the comments. Please give it another look.

wmorgan (approved):
That works for me! Thank you.

PR #955, opened 2021-03-02
Update protocol detection docs with opaque ports

adleong:

### DIFF
--- a/linkerd.io/content/2/features/protocol-detection.md
+++ b/linkerd.io/content/2/features/protocol-detection.md
@@ -31,37 +31,54 @@ Linkerd cannot automatically recognize the protocol used on the connection.
 (Note that TLS-enabled connections work as normal, because TLS itself is a
 client-speaks-first protocol.)
 
-The following protocols are known to be server-speaks-first:
+The following protocols are known to be server-speaks-first and are treated as
+opaque by default:
 
 * 25   - SMTP
 * 3306 - MySQL
 * 4222 - NATS
-* 27017 - MongoDB
+* 5432 - PostgreSQL
+* 11211 - Memcached (clients do not issue any preamble, which breaks detection)
 
 If you're working with a protocol that can't be automatically recognized by
-Linkerd, use the `--skip-inbound-ports` and `--skip-outbound-ports` flags when
-running `linkerd inject`.
+Linkerd, you will need to set the `config.linkerd.io/opaque-ports` annotation on
+both the Pod template spec of the workload and on the Service.  This annotation
+tells Linkerd to skip protocol detection and immediately treat connections on
+those ports as opaque TCP.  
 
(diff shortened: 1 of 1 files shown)


dadjeibaah (approved):
Reviewed mostly the copy. This LGTM! 📦 🚢

linkerd.io/content/2/features/protocol-detection.md
```diff
@@ -31,37 +31,54 @@ Linkerd cannot automatically recognize the protocol used on the connection.
...
 
-For example, if your application makes requests to a MySQL database running on
-port 4406, use the command:
+Setting this annotation on the Service resource tells meshed clients to skip
+protocol detection when proxying connections on those ports to the Service.
+Similarly, setting this annotation on the Pod template spec tells Linkerd to skip
+protocol detection when reverse-proxying incoming connections on those ports.
+This means that it is still important to set this annotation on Services that use
+server-speaks-first protocols if they have any meshed clients, even if that
+Service itself is not meshed.
```
kleimkuhler, line 55: The first part of this explanation is true for pod resources as well right? Meshed clients will skip protocol detection when proxy connections on those ports to the pod.

The reverse-proxying is pod specific though so the second part of this is good as-is.

adleong: Yeah, that's true.  I'll try to think of a way to work that in without making this more confusing.  Calling pods directly is less common than using services so I think it's a more advanced case.

linkerd.io/content/2/features/protocol-detection.md
```diff
@@ -31,37 +31,54 @@ Linkerd cannot automatically recognize the protocol used on the connection.
...
-The following protocols are known to be server-speaks-first:
+The following protocols are known to be server-speaks-first and are treated as
+opaque by default:
 
 * 25   - SMTP
 * 3306 - MySQL
 * 4222 - NATS
-* 27017 - MongoDB
+* 5432 - PostgreSQL
+* 11211 - Memcached (clients do not issue any preamble, which breaks detection)
```
wmorgan, line 41: Late to this review, but I don't see 4222 anywhere in the code. Also shouldn't 587 be in here (another SMTP port)? 

Finally, I think that Postgres wasn't actually a SSF protocol, it was just one that we handled wrong for some other reason.

kleimkuhler: Yep this can be updated. The default list is: `25,443,587,3306,5432,11211`. `4222` is not in there and `587` is.

The reasoning behind keeping `5432` a default opaque port is that we already know it's not HTTP, so there is no benefit to making it not-opaque.

kleimkuhler: #998

PR #984, opened 2021-03-11
Blog post edit

PaulMichaelArmstrong:
Signed-off-by: Paul Armstrong <paul@buoyant.io>


wmorgan:
I think we need additional wiring before this field actually hits the meta tag.

PaulMichaelArmstrong:
Yeah, I was just seeing it didn't work. :(

PR #988, opened 2021-03-12
Update container

mikutas:
with the same image used in validate workflow


kleimkuhler:
@tkms0106 It looks like there are other places where we have `gcr.io/linkerd-io/website-builder`. Additionally, it looks like there is a newer `1.3.3` version on docker/buoyantio.

We should update the additional locations and do a version bump while we're changing them.

mikutas:
recreated #1009

PR #975, opened 2021-03-05
feat(ci): add markdown seo check

PatrickHeneise:
Signed-off-by: Patrick Heneise <patrick@zentered.co>

PR #995, opened 2021-03-14
docs(traefik): add note about default ports

rio:
Traefik starts it's TLS port on 8443 by default but exposes it with a service on 443. The linkerd proxy doesn't include it in it's default ports, this explains how to solve it.

Signed-off-by: Rio Kierkels <riokierkels@gmail.com>


rio:
The SEO validation job got added 2 days ago through, #975 by @PatrickHeneise. Maybe Patrick can shed some light on why it's failing. It just seems to not find what it's looking for. Disclaimer, I have no experience with SEO.

```
 Starting Markdown SEO Check ...
No valid frontmatter in linkerd.io/content/2.10/tasks/using-ingress.md
Error: HttpError: Resource not accessible by integration
Error: Resource not accessible by integration
```

PatrickHeneise:
Thanks for the heads up. There's an issue with PRs from forks. Please ignore the action for now.

rio:
Thanks @PatrickHeneise, Just waiting for the review.

linkerd.io/content/2.10/tasks/using-ingress.md
```diff
@@ -322,6 +322,28 @@ spec:
...
+*non-root* the traefik process has no privileges to open ports below `1024`.
+Adding port `8443` to the opaque ports will restore functionality, just remember
+that setting opaque ports will remove the defaults. You can add this in the
+`values.yaml` file like this:
+
+```yaml
+deployment:
+  podAnnotations:
+    linkerd.io/inject: enabled
+    config.linkerd.io/opaque-ports: "8443"
```
kleimkuhler, line 342: Thanks for documenting this @rio! Looking at `values.yaml` it looks like `podAnnotations` is at the root; there is no `deployment` before it.

rio: Any time!

But are we looking at the same chart? This is the exactly line I pulled it from: https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L22

kleimkuhler: Ah yes sorry about that; that's correct.

kleimkuhler (dismissed):

linkerd.io/content/2.10/tasks/using-ingress.md
```diff
@@ -322,6 +322,28 @@ spec:
...
+creates a websecure entrypoint on port `8443` to handle TLS traffic but exposes
+it using a service on port `443`. Because the linkerd-proxy operates on a pod
+level and not a service level, the default set of opaque ports will not work as
+it does not include port `8443`. This results in protocol detection timeouts
+visible in the linkerd-proxy container logs. You could tell the traefik
+container to listen on the actual port `443` but since that container runs as
+*non-root* the traefik process has no privileges to open ports below `1024`.
+Adding port `8443` to the opaque ports will restore functionality, just remember
+that setting opaque ports will remove the defaults. You can add this in the
+`values.yaml` file like this:
```
kleimkuhler, line 336: @rio One more thing. Reading this could be a little confusing to users not familiar with the annotation.

Adding the annotation like you suggest below does not override the default list of opaque ports; it simply marks the pods with these annotations as opaque workloads on `8443`.

To actually override the default list would require setting that during install or upgrade with the `--set proxy.opaquePorts` flag.

For example, if I wanted to override the default list I would do `linked install --set proxy.opaquePorts="8443"`. If I wanted to add `8443` and just not worry about additional annotations, I would append to [default list](https://github.com/linkerd/linkerd2/blob/main/charts/linkerd2/values.yaml#L107): `linkerd install --set proxy.opaquePorts="25,443,587,3306,5432,11211,8443"`.

---

So for the suggestion here, I would recommend you list the annotation option first since we're assuming a user already has installed Linkerd.

Then, mention that installing and see `8443` as opaque is another work around.

In general we can improve documentation on this configuration, but I think this is a good place to start. Thanks again for taking care of this!

rio: I understand what you're getting at but just to clear it up between us: the default ports setup during install system wide are not overwritten at a system level. Only the defaults for that deployment are overwritten. For example here is the environment variable for the linkerd proxy without the opaque ports annotaition set:

```
LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION:  25,443,587,3306,5432,11211
```
The ports listed there are the default ports set during install. So far so good. Now, setting the annotation as described on another deployment will result in this environment variable set:

```
LINKERD2_PROXY_INBOUND_PORTS_DISABLE_PROTOCOL_DETECTION:  8443
```
Here you see the defaults have been replaced for this deployment.

Now that we've cleared that up I'll try to word it a bit better tomorrow as requested :)

kleimkuhler: Ah yes I was misunderstanding your original wording. Thanks for clarifying! So yes, we do override the opaque ports for this specific pod. What I was answering is that we don't override the opaque ports for the whole installation.

Thanks for catching the miscommunication there and I'll look out for a change tomorrow!

rio:
@kleimkuhler, I've reworded it but it has gotten a bit longer. Refactoring the docs for all these ingress controllers might be needed at some point.

alpeb:
@rio protocol detection timeouts only occur on server-speaks-first protocols, which is not the case of TLS, so port 8443 shouldn't require to be opaque as the proxy would immediately detect it's not http and let it through. Nevertheless, I could replicate this issue in my tests in 2.10, but not in the [latest edge](https://github.com/linkerd/linkerd2/releases/tag/edge-21.3.3). Would you mind giving it another try yourself with that edge and let us know how it goes?

rio:
@alpeb, I was a bit surprised that it didn't work as I did understand it was server-speak-first protocols that don't get detected properly. However if it should just work why is port 443 set as an opaque port by default?

Anyway the edge-21.3.3 upgrade didn't fix it for me. I still have to set 8443 as an opaque port. Could it be architecture related maybe? I'm running on arm64.

rio:
I've made an issue in the linkerd2 repo with a script that can reproduce this locally with both stable and edge on either k3d and kind. https://github.com/linkerd/linkerd2/issues/5948

kleimkuhler:
@rio Thanks for opening the issue on this and helping us figure out what the actual issue was—really appreciate it! This should be good to close now that it's tracked in linkerd/linkerd2#5948 and the fix will be included in this week's edge.

rio:
Yes no problem, I wanted to actually wait until it's landed in a release but I guess the docs aren't needed either way. Closing.

PR #998, opened 2021-03-17
Update default opaque ports list

kleimkuhler:
Signed-off-by: Kevin Leimkuhler <kevin@kleimkuhler.com>


kleimkuhler:
It seems the SEO check is failing this, so I think after an approval we can just admin merge while that gets fixed.

cpretzer (approved):
LGTM

PR #1005, opened 2021-03-19
fix top demo

FingerLiu:
this do not work:
```bash
linkerd -n linkerd viz top deploy/linkerd-web
```

this works:
```bash
linkerd -n linkerd-viz viz top deploy/web
```


kleimkuhler (approved):
Thanks @FingerLiu!

PR #1010, opened 2021-03-23
fix: meta descriptions & summaries

PatrickHeneise:


cpretzer (approved):
The descriptions are displaying correctly in the meta tags 🚢

PR #1013, opened 2021-03-23
Update the edge install script to download M1 CLI

dadjeibaah:
With the addition of linkerd/linkerd2#5907, users with the new MacOS M1
arch need to be able to easily install the CLI. This change modifies the
edge bash script to detect the new arch and install the correct image.


alpeb (approved):
LGTM :+1:

PR #1022, opened 2021-04-02
Update Google Analytics Tag

cpretzer:
* Changed `<script>` element in `meta.html` to use new Google Global Tag


cpretzer:
Closing this for now because we want to continue to support third-party tags

PR #1033, opened 2021-04-09
Fix two typos in distributed tracing

sudomakeinstall2:


cpretzer (approved):
thanks for the corrections @sudomakeinstall2

PR #1046, opened 2021-04-17
Document new checks added as part of #5943

Szymongib:
Related PR https://github.com/linkerd/linkerd2/pull/5943

Related Issue https://github.com/linkerd/linkerd2/issues/5718

Signed-off-by: Szymon Gibała <szymongib@gmail.com>

### DIFF
--- a/linkerd.io/content/2.10/tasks/troubleshooting.md
+++ b/linkerd.io/content/2.10/tasks/troubleshooting.md
@@ -1073,9 +1073,8 @@ Example failure:
         linkerd.io/inject
 ```
 
-Setting `linkerd.io/inject`, `config.linkerd.io/*` or
-`config.alpha.linkerd.io/*` as pod labels does not have any effect on
-the service mesh but it can have when set as annotations.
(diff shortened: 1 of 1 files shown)


kleimkuhler (commented):
Thanks for updating this @Szymongib! I think the wording could change a little; saying these labels or annotations _do_ have an effect is a little more accurate than the current _can_ have an effect.

So I'm thinking the following:
```
Setting `linkerd.io/inject`, `config.linkerd.io/*` or
 `config.alpha.linkerd.io/*` as pod labels does not have any effect on
 the service mesh but it can have when set as annotations.
```

Would be worded as:
```
`linkerd.io/inject`, `config.linkerd.io/*` or `config.alpha.linkerd.io/*` should
be annotations in order to take effect.
```

Szymongib:
Thanks @kleimkuhler, your suggestion does sound better, I have adjusted the PR.

kleimkuhler (approved):
Awesome thanks @Szymongib

PR #1057, opened 2021-04-27
add more opaque/skip ports docs, and a protocol table

wmorgan:
Try and clarify skip-ports and opaque-ports based on some recent Slack questions.

### DIFF
--- a/linkerd.io/content/2.10/features/protocol-detection.md
+++ b/linkerd.io/content/2.10/features/protocol-detection.md
@@ -7,61 +7,82 @@ aliases = [
 ]
 +++
 
-Linkerd is capable of proxying all TCP traffic, including TLS'd connections,
+Linkerd is capable of proxying all TCP traffic, including TLS connections,
 WebSockets, and HTTP tunneling.
 
-Linkerd performs *protocol detection* to determine whether traffic is HTTP or
-HTTP/2 (including gRPC). If Linkerd detects that a connection is using HTTP or
-HTTP/2, Linkerd will automatically provide HTTP-level metrics and routing
-without configuration from the user. (See
-[HTTP, HTTP/2, and gRPC Proxying](../http-grpc/) for more.)
+In most cases, Linkerd can do this without configuration. To do this, Linkerd
+performs *protocol detection* to determine whether traffic is HTTP or HTTP/2
+(including gRPC). If Linkerd detects that a connection is HTTP or HTTP/2,
+Linkerd will automatically provide HTTP-level metrics and routing.
 
-If Linkerd *cannot* determine that a connection is using HTTP or HTTP/2, Linkerd
-will proxy the connection, but will only be able to provide byte-level metrics.
-Note that this also applies to TLS'd HTTP connections if the application
-initiates the TLS, as Linkerd will not be able to observe the HTTP transactions
-in this connection.
+If Linkerd *cannot* determine that a connection is using HTTP or HTTP/2,
+Linkerd will proxy the connection as a plain TCP connection, applying
+[mTLS](../automatic-mtls/) and providing byte-level metrics as usual.
 
-## Configuring Protocol Detection
+{{< note >}}
+Client-initiated HTTPS will be treated as TCP, not as HTTP, as Linkerd will not
+be able to observe the HTTP transactions on the connection.
+{{< /note >}}
 
-In some cases Linkerd's protocol detection requires configuration. Currently,
-this is required for unencrypted "server-speaks-first" protocols, or protocols
-where the server sends data before the client sends data. In these cases,
-Linkerd cannot automatically recognize the protocol used on the connection.
-(Note that TLS-enabled connections work as normal, because TLS itself is a
-client-speaks-first protocol.)
+## Configuring protocol detection
 
-The following protocols are known to be server-speaks-first or non-HTTP
-protocols and are treated as opaque by default:
+In some cases, Linkerd's protocol detection requires configuration to avoid a
+10-second connection delay. This configuration is typically required when using
+certain "server-speaks-first" protocols, or protocols where the server sends
+data before the client does, to communicate with off-cluster services.
 
-* 25,587 - SMTP
-* 443 - HTTPS
-* 3306 - MySQL
-* 5432 - PostgreSQL
-* 11211 - Memcached: clients do not issue any preamble, which breaks detection
+{{< note >}}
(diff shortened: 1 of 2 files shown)


linkerd.io/content/2.10/features/protocol-detection.md
```diff
@@ -7,61 +7,82 @@ aliases = [
...
+
+## Configuring protocol detection
+
+In some cases, Linkerd's protocol detection requires configuration to avoid a
+10-second connection delay. This configuration is typically required when using
+certain "server-speaks-first" protocols, or protocols where the server sends
+data before the client does, to communicate with off-cluster services.
+
+{{< note >}}
+Regardless of the underlying protocol, client-initiated TLS connection do not
```
kleimkuhler, line 35:
```suggestion
Regardless of the underlying protocol, client-initiated TLS connections do not
```

linkerd.io/content/2.10/features/protocol-detection.md
```diff
@@ -7,61 +7,82 @@ aliases = [
...
+
+## Marking a port as opaque
+
+The `config.linkerd.io/opaque-ports` annotation tells Linkerd to skip protocol
+detection for a specific set of ports, and immediately treat connections on
+those ports as TCP. This is useful for bypassing protocol detection while still
+proxying the connection.
+
+The opaque-ports annotation can be set on the Pod template spec of a workload
+or on a Service resource. Setting it on a Service resource tells meshed clients
```
kleimkuhler, line 71: It may be worth mentioning it can also be set on a namespace and then all services and workloads within that namespace will inherit the annotation.

linkerd.io/content/2.10/tasks/adding-your-service.md
```diff
@@ -69,22 +71,33 @@ If everything was successful, you'll see `linkerd-proxy` in the output, e.g.:
...
+
+While the proxy starts very quickly, Kubernetes doesn't provide any guarantees
+about container startup ordering, so the application container may start before
+the proxy is ready. This means that any connections made immediately at app
+startup time may fail until the proxy is active.
+
+In many cases, this can be ignored: the application will ideally retry the
+connection, or Kubernetes will restart the container after it fails, and
+eventually the proxy will be ready. Alternatively, you can use
+[linkerd-await](https://github.com/linkerd/linkerd-await) to delay the
```
kleimkuhler, line 84: So in order to use `linkerd-await` users will need to upgrade to a recent edge. On the recent edges, this feature is enabled by default so they don't actually need to do anything after upgrading. It may be worth wording it in this way:
```suggestion
eventually the proxy will be ready. Alternatively, you can upgrade to a recent
edge that has the [linkerd-await](https://github.com/linkerd/linkerd-await)
feature enabled by default—used to delay the application container until the
proxy is ready. Or, ...
```

linkerd.io/content/2.10/features/protocol-detection.md
```diff
@@ -7,61 +7,82 @@ aliases = [
...
+
+There are two basic mechanisms for configuring protocol detection: _opaque
+ports_ and _skip ports_. Marking a port as _opaque_ instructs Linkerd to skip
+protocol detection and proxy the connection as a TCP stream. Marking a port as
+_skip_  bypasses the proxy entirely. Opaque ports are generally preferred (as
+it allows Linkerd to provide mTLS, TCP-level metrics, etc), but crucially,
+opaque ports can only be used for services on the cluster.
+
+The following table summarizes some common protocols and the configuration
+necessary to handle them. The "on-cluster config" column refers to the
```
kleimkuhler, line 48: It may be worth mentioning that some common protocols and their standard ports are marked as opaque by default, which is why this table says that no configuration is needed for these protocols.

kleimkuhler (commented):
Looks good! Left a few comments.

linkerd.io/content/2.10/features/protocol-detection.md
```diff
@@ -7,61 +7,82 @@ aliases = [
...
+configuration when the destination is *on* the same cluster; the "off-cluster
+config" to when the destination is external to the cluster.
+
+| Protocol | Default port(s) | On-cluster config | Off-cluster config|
+|----------|-----------------|---------------------------|----------------------------|
+| SMTP | 25, 587 | none\* | skip ports |
+| MySQL | 3306 | none\* | skip ports |
+| PostgreSQL | 5432 | none\* | skip ports |
+| Memcache | 11211 | none\* | skip ports |
+| ElasticSearch | 9300 | opaque ports | skip ports |
```
christianhuening, line 58: We have port `6379` for Redis in our default skip list in `2.9.4` and would move it to the opaque ports in `2.10.x`

electrical: Mysql with Galera replication.  I don't know what's required for off-cluster but I presume it all should get skipped.
on-cluster:
2.9
* Skip ports: 3306, 4444, 4567, 4568
2.10
* Opaque ports: 3306, 4444
* Skip ports: 4567, 4568

wmorgan: @electrical thanks! Can 4567 and 4568 not be opaque ports as well?

linkerd.io/content/2.10/tasks/adding-your-service.md
wmorgan: Not sure I understand. We've been using linkerd-await for years. What is different about recent edge releases?

kleimkuhler: @wmorgan Yep this is good as is. I read this thinking you were referencing the `config.linkerd.io/proxy-await` annotation that was recently added to edges (linkerd/linkerd2#5967), but that will not be part of 2.10. So, this is good as is.

kleimkuhler (approved):
👍

linkerd.io/content/2.10/features/protocol-detection.md
electrical: @wmorgan I'm afraid not after extensive testing. This is due to how Galera replication uses the source IP for keeping track of the clients. When it goes through the proxy everything comes from 127.0.0.1 and loses track of it. 
More details of that are in https://github.com/linkerd/linkerd2/issues/5913

linkerd.io/content/2.10/tasks/adding-your-service.md
wmorgan: Gotcha. Yeah I am very excited for that annotation.

PR #1063, opened 2021-04-30
Fix latest stable version number retrieval

alpeb:
Followup to #1060

The problem with the previous way was that we just picked the latest (by
date) version published in Github, which doesn't work when we publish a
minor version for an older major version. E.g. the latest (by version)
stable being 2.10.1, then we publish 2.9.5, then this algorithm will
wrongly pick 2.9.5 as the latest stable.

In this fix we get the list of stable versions from github, sort using
`sort -V` and pick the latest item.

### DIFF
--- a/linkerd.io/content/2.10/features/protocol-detection.md
+++ b/linkerd.io/content/2.10/features/protocol-detection.md
@@ -7,61 +7,90 @@ aliases = [
 ]
 +++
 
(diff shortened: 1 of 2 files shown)


dadjeibaah (approved):
Tested and works well! Thanks for fixing this 📦 🚢

PR #1065, opened 2021-05-04
Fix load balancing link

obrienrobert:
For both 2.9 and 2.10, the `its own EWMA load balancing` link within the https://linkerd.io/2.9/tasks/using-ingress/ and https://linkerd.io/2.10/tasks/using-ingress/ doc is dead.

### DIFF
--- a/linkerd.io/content/2.10/tasks/using-ingress.md
+++ b/linkerd.io/content/2.10/tasks/using-ingress.md
@@ -10,7 +10,7 @@ can be run with your Ingress Controller.
 
 When the ingress controller is injected with the `linkerd.io/inject: enabled`
 annotation, the Linkerd proxy will honor load balancing decisions made by the
-ingress controller instead of applying [its own EWMA load balancing](https://linkerd.io../../features/load-balancing/).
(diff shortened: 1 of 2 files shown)


cpretzer (changes requested):
Thanks for helping with this @obrienrobert

The link validation check reports that the URL is invalid, and the relative URLs for the links don't need the `..`

Can you try setting the URL to `/features/load-balancing` ?

cpretzer:
Thanks for the quick fix @obrienrobert It looks like some of our redirect rules cause the link to go to the linkerd 1 docs.

I'll have a look to figure out what the correct relative path is for these pages. I'll put an update here when I sort it out.

obrienrobert:
@cpretzer no problem. Please feel free to close this PR if you think that a new PR is required to fix any additional issues with the links. It's just something I came across when reading through the docs.

cpretzer:
@obrienrobert thanks again for this PR. I found a few more malformed links, and updated them in #1066 

I'm going to close this PR, and please keep up the great work on making the docs awesome. We're looking forward to your next PR!

PR #1069, opened 2021-05-07
New versioned `linkerd check` URL hints target

alpeb:
Added a new directory `/linkerd.io/content/2.10/checks/` that takes care of redirecting the new URL hints format to be adopted in linkerd/linkerd2#6102

So from 2.10 onwards, `linkerd check` and friends will output versioned URLs, as instructions may vary depending on the version.

For example `https://linkerd.io/2.10/checks#pre-k8s` will get redirected to `https://linkerd.io/2.10/tasks/troubleshooting/#pre-k8s`.

Edge and developer versions will have hints using version `2`, like `https://linkerd.io/2/checks#pre-k8s`, which a Cloudflare rule automatically redirects to the latest version, in this case `http://linkerd.io/2.10/checks#pre-k8s` which itself will get redirected to `http://linkerd.io/2.10/tasks/troubleshooting/#pre-k8s`.

The old unversioned URL hints such as `http://linkerd.io/checks/#pre-k8s` still get redirected to `https://linkerd.io/2.10/tasks/troubleshooting/#pre-k8s` as per the redirection in `/linkerd.io/static/checks/`.

Note that javascript redirection was necessary because I couldn't make Hugo aliases follow the anchor fragment.


cpretzer (approved):
🚢

PR #1071, opened 2021-05-07
New glossary terms

CathPag:


CathPag:
Whoever merges this, could you please verify the links to the Linkerd docs of the reliability term are correct?

linkerd.io/content/service-mesh-glossary/_index.md
```diff
@@ -165,9 +197,33 @@ support provides seamless and secured communication across clusters, in a
...
+
+## Reliability
+
+Reliability is a system property that measures how well the system
+responds to failure. The more reliable a system is, the better it
+can handle individual components being down or degraded. For
+multi-service or microservice applications, a service mesh can
+be used to increase the reliability by applying[techniques
+like retries and
+timeouts](https://linkerd.io/2.10/features/retries-and-timeouts/)
```
wmorgan, line 217:
```suggestion
timeouts](/2/features/retries-and-timeouts/)
```

linkerd.io/content/service-mesh-glossary/_index.md
```diff
@@ -165,9 +197,33 @@ support provides seamless and secured communication across clusters, in a
...
+to cross-service calls, by [load balancing
+in intelligent ways](https://linkerd.io/2.10/features/load-balancing/),
```
wmorgan, line 219:
```suggestion
in intelligent ways](/2//features/load-balancing/),
```

linkerd.io/content/service-mesh-glossary/_index.md
```diff
@@ -165,9 +197,33 @@ support provides seamless and secured communication across clusters, in a
...
+by [shifting traffic](https://linkerd.io/2.10/features/traffic-split/)
```
wmorgan, line 220:
```suggestion
by [shifting traffic](/2//features/traffic-split/)
```

wmorgan (commented):
Links should be relative

cpretzer:
The [validation checks](https://github.com/linkerd/website/pull/1071/checks?check_run_id=2567728391#step:5:37) indicate that the URLs for `/2/...` don't exist and this is correct because the redirect is not part of the validation, as far as I can tell. For example, if you go directly to https://linkerd.io/2/features/automatic-mtls, the page will redirect to https://linkerd.io/2.10/features/automatic-mtls.

This doesn't appear to work in the context of these validation checks, which requires some investigation.

In the short term, I think we can update the glossary links to go directly to the latest version, but the long term fix will require more effort because we don't want to have to update these links for every new release of Liinkerd.

cpretzer (changes requested):
See note in the PR about using the latest version in the URLs

wmorgan:
Weird, we have these /2/ links all over the place, don't we? Why would the validation fail just for these?

cpretzer:
I'll dig into it tomorrow

CathPag:
@cpretzer just following up on this. Could we get it live early this week?

cpretzer (changes requested):

cpretzer:
> Weird, we have these /2/ links all over the place, don't we? Why would the validation fail just for these?

We do have `/2/` links in other places, so this is unexpected. Once the URLs are cleaned up, I'll merge this and fix it from a separate branch.

If you go to https://linkerd.io/2/features/load-balancing (or any of the URLs) the redirects work. So, it seems like there is something going awry with the CI checks and the redirects, but I'd expect all the `/2/` links to fail

cpretzer (approved):
The changes look good, despite the failing CI checks. 

To expedite getting this content out, I'm going to merge this and create a new branch to update the links.

PR #1083, opened 2021-05-24
Add HTTPS to linkerd CLI install instructions

christophetd:
Seems to be the only place in the docs where a plain HTTP link remains. It seems quite important to have HTTPS since the command is downloading a bash script and piping it to bash.


cpretzer (approved):
thanks @christophetd !

PR #1091, opened 2021-06-03
Referenced Buoyant article on Emissary.

JasonMorgan:
Wanted to reference the Buoyant article on integrating with Emissary as it provides the same information with more detail.


JasonMorgan:
Wrong branch.

PR #1092, opened 2021-06-03
Referenced Buoyant article on Emissary.

JasonMorgan:
Wanted to reference the Buoyant article on integrating with Emissary as it provides the same information with more detail.

This time using the correct branch ;)

### DIFF
--- a/linkerd.io/content/2.10/tasks/using-ingress.md
+++ b/linkerd.io/content/2.10/tasks/using-ingress.md
@@ -431,6 +431,11 @@ You can then use this IP with curl:
 curl -H "Host: example.com" http://external-ip
 ```
 
+{{< note >}}
+You can also find a more detailed guide for using Linkerd with Emissary Ingress,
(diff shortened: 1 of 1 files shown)


cpretzer:
Should we just update the doc with the content?

We should link to the Buoyant article, just like we do with other tutorials like flagger. Since Ambassador is changing to Emissary, it makes sense to update the entire section in the linkerd docs

JasonMorgan:
I wouldn't change the name in the official docs yet. It looks like renaming Ambassador API Gateway to Emissary Ingress is going to be a WIP for awhile.

cpretzer:
> I wouldn't change the name in the official docs yet. It looks like renaming Ambassador API Gateway to Emissary Ingress is going to be a WIP for awhile.

Gotcha, thanks for clarifying @JasonMorgan

cpretzer (approved):
LGTM nice work @JasonMorgan :shipit:

PR #1094, opened 2021-06-03
Clarify Canary Release Tutorial

alpeb:
People (and I) were trying to test things by port-forwarding `podinfo`,
but that directs you straight into the primary pod, skipping the
trafficsplit logic. Towards the end of the page there were proper
instructions to use the frontend service instead, but it's clearer if we
do that from the beginning.

This was clarified after this convo with a user:
https://linkerd.slack.com/archives/C89RTCWJF/p1622732904469200


cpretzer (approved):
LGTM 🚢

PR #1097, opened 2021-06-08
Add instructions for using helm on "Manually Rotating Control Plane TLS Credentials" page

mikutas:


cpretzer (approved):
works great! thanks @mikutas

PR #1123, opened 2021-07-09
Update Contour instructions

alpeb:
Fixes linkerd/linkerd2#6237

- Since a few releases ago, `linkerd inject` will fail outright if
`automountServiceAccountToken` is not `true`. Reshuffled the instructions to
account for that.
- Added note about adding the
`config.linkerd.io/skip-outbound-ports: 8001` annotation when in ingress mode.
- Removed the `spec.routes.services.namespace` field from the sample
`HTTPProxy` resource, which doesn't appear in the resource API doc.
- Replaced `127.0.0.1.xip.io` with `127.0.0.1.nip.io` as the former is down.


cpretzer (approved):
lgtm

cpretzer:
@alpeb I'm going to verify these steps and I'll get this merged

PR #1128, opened 2021-07-15
simplified method to add proxy headers with nginx ingress controller

dkulchinsky:


cpretzer (approved):
Thanks for this @dkulchinsky

PR #1129, opened 2021-07-16
Upgrade Ingress resources

alpeb:
Ref linkerd/linkerd2#6508

Ingress version `extensions/v1beta1` will disappear in k8s 1.22, so
this upgrades it to `networking.k8s.io/v1` alongside with other required
Ingress changes.


cpretzer (approved):
lgtm 🚢

PR #1134, opened 2021-07-29
Fix firewall description

jsushetski:
Fixed the ports in the description of the firewall rule required for private GKE clusters.


kleimkuhler (approved):
Thanks @jsushetski

PR #1135, opened 2021-07-30
Add docs for multicluster with headless services

mateiidavid:
* Added a new "/tasks/" page with a walkthrough on how to use multi-cluster w/ StatefulSets (note: the cli version suggested here is 2.11, if we release these docs before 2.11 I can change it into the most recent edge)
* Added an explanation on how we mirror headless services for multi-cluster in "/features/multicluster/"

Feedback more than welcome! I have included a demo repo instead of creating clusters and installing Linkerd through shell commands. K3d is a bit hard to bootstrap so thought this would save some time. I'm not sure at the moment if I forgot to update anything, off the top of my head this should be it.

Let me know what you think, more than happy to make any changes. At the moment, the feature is not completed, we still need to add the `multicluster link` flag.


linkerd.io/content/2.10/tasks/multicluster-using-statefulsets.md
```diff
@@ -0,0 +1,332 @@
...
+</body>
+</html>
+```
+
+As you can see, we get the same response back! But, nginx is in a different
+cluster. So, what happened behind the scenes?
+
+  1. When we mirrored the headless service, we created a clusterIP service for
+     each pod. Since services create DNS records, naming each endpoint with the
+     hostname from the target gave us these pod FQNS
```
cpretzer, line 297: Should this be `FQDN` instead of `FQDS`?

mateiidavid: Oops, yup! `FQDN` indeed.

mateiidavid:
> LGTM
> 
> Should this be merged before or after the `2.11` release?

Thanks for the review! From my perspective, it's good to go as soon as https://github.com/linkerd/linkerd2/issues/6562 is done. Currently, the feature is in the latest edge release but it needs to be turned on, the only way to do it is by editing the `linkerd multicluster link` output manually. Once the flag is in, I think we're good to release the documentation (we might expect users to experiment with it before 2.11), with the mention that the linkerd version needs to be edited in the "Prerequisites" section.

cpretzer (approved):

cpretzer (approved):
LGTM!

Great tutorial @mateiidavid 

I think we could probably refer to the multicluster setup docs instead of adding this set of instructions, but that change can be made in a different PR

PR #1137, opened 2021-08-01
Update _index.md

Hiiirad:
Change "incubating project" to "graduated project" :)


wmorgan:
Good catch. Thanks @Hiiirad ! 

We can skip the DCO check for this one.

cpretzer (approved):
🎓 🚢

PR #1141, opened 2021-08-03
docs: typo gloval to global

jef:
## Description

Typo in documentation. Should be `global` instead of `gloval`.


cpretzer (approved):
Thanks @jef good catch 🚢

PR #1142, opened 2021-08-03
Fix double forward slash issue in extensions json

dadjeibaah:
This change fixes an issue where the screenshot link for the
extensions/index.json endpoint would contain a double forward slash.
This was caused by the `list.json.json` adding a forward slash
regardless of whether the base URL did or did not already contain a
forward slash. This change now premptively strips a forward slash from
the base URL if it exists and then formats the screenshot URL correctly.


cpretzer (approved):
Good catch @dadjeibaah 

I didn't notice this extra slash when I checked the images

dadjeibaah:
@cpretzer, yea same here, it did show up once I looked at it live https://linkerd.io/extensions/index.json

PR #1149, opened 2021-08-09
Fix broken links in install scripts, add redirect

siggy:
The stable and edge install scripts were referencing a broken link:
https://linkerd.io/2.10/next-steps

Modify the scripts to instead reference:
https://linkerd.io/2/tasks

Also, introduce a redirect from `/next-steps` to `/tasks`.

### DIFF
--- a/linkerd.io/content/2.10/tasks/_index.md
+++ b/linkerd.io/content/2.10/tasks/_index.md
@@ -1,6 +1,9 @@
 +++
 title = "Tasks"
 weight = 4
+aliases = [
+  "./next-steps/"
+]
 +++
 
 As a complement to the [Linkerd feature docs]({{% ref "../features" %}}) and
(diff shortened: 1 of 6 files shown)


cpretzer (commented):
There are a couple of things failing, but I don't think they're related to this change and this code looks good

1. The [source file check](https://github.com/linkerd/website/pull/1149/checks?check_run_id=3282654201) is failing. And the error is in the github action when it's trying to list file. 
2. The redirect isn't working. After some debugging, it looks like any of the redirects with `-` in the name is failing locally, on the preview, and on the live site. For example https://linkerd.io/2.10/tap-rbac and https://linkerd.io/2.10/dns-rebinding result in 404s, while https://linkerd.io/2.10/upgrade and https://linkerd.io/2.10/check both work

Both are going to require some debugging

cpretzer (approved):
LGTM, thanks @siggy

PR #1151, opened 2021-08-11
chore: update markdown check

PatrickHeneise:
there were some breaking changes in @actions/github that I missed. `pull_request` should suffice here.


cpretzer (approved):
lgtm, thanks @PatrickHeneise

PR #1175, opened 2021-09-05
Add tip when running linkerd on a remote cluster

juampynr:
After debugging why I was getting `ERR_CONNECTION_REFUSED` at the Linkerd's viz dashboard I found out that in order to access it externally I need to set the `--address=0.0.00` option and add adjust my hosts file.


cpretzer:
@juampynr there are a few comments here that your feedback will be helpful for. After that we can merge this. 😄

cpretzer:
Closing due to inactivity

PR #1179, opened 2021-09-06
Marketing tags cleanup

alpeb:
- Removed Universal Analytics tag `UA-60040560-2`
- Replaced Google Tag Manager tag `GTM-KHF5SSN` with `GTM-PXBDGKJ` which
  is solely for linkerd.io and is not shared with buoyant.io. Also note
  this new tag will add the tag `UA-60040560-2` automatically.


wmorgan (approved):
Ty!

PR #1184, opened 2021-09-10
Add more iptables docs

mateiidavid:
Unfortunately, for the table reference, I need to keep everything at col. 80 to avoid the md length check. This means I can't use very descriptive descriptions, tried to be as accurate as I can be.


kleimkuhler:
@mateiidavid You should be able to wrap those sections with:
```
<!-- markdownlint-disable MD013 -->
...
<!-- markdownlint-enable MD013 -->
```

[MD013](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md013---line-length) is the line length lint for the markdown linter that we use.

mateiidavid:
@kleimkuhler wow that's neat! I've pushed an update with more descriptive explanations. I've also added the actual iptables command as a column.

cpretzer (approved):
🚢

PR #1191, opened 2021-09-22
Change iptables documentation

mateiidavid:
Updated with diagrams and text.


mateiidavid:
@adleong I kept the tables in but I'm really not opposed to removing them

alpeb:
Thanks for the writeup, this is precious info!
I'm still going through it, but one thing I noticed is that clicking on the diagrams brings them upfront, but it looks like they have a transparent background so they don't show up right. A white background should work better in this case.

linkerd.io/content/2.11/reference/iptables.md
```diff
@@ -8,13 +8,172 @@ container`](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)
...
+target the outbound port. The proxy will process the packet and then forward it
+to its destination. The new packet will take the same path through the `OUTPUT`
+chain, however, it will stop at the first rule, since it was produced by the
+proxy.
+
+The substantiated explanation applies to a packet whose destination is another
+service, outside of the pod. In practice, an application can also send traffic
+locally. As such, there are two other possible scenarios that we will explore:
+_when a service talks to itself_ (by sending traffic over localhost or by using
+its own endpoint addresS), and when _a service talks to itself through a
```
alpeb, line 108:
```suggestion
its own endpoint address), and when _a service talks to itself through a
```

mateiidavid:
Rebased off main and committed your suggestions @alpeb.

Good point about the diagrams, I didn't consider that. I'll have to re-do them but hopefully that won't be too much work.

cpretzer (approved):
great work @mateiidavid 🚢

PR #1194, opened 2021-09-27
Fix doc about disabling dashboard host header check

alpeb:
Fixes linkerd/linkerd2#6969

Using an empty string for `enforceHostRegexp` doesn't disable that
check; it just falls back to the default of checking against the default
`^(localhost|127\.0\.0\.1|web.linkerd-viz.svc.cluster.local|web.linkerd-viz.svc|\[::1\])(:\d+)?$`.
So we just fix the docs saying if you wanna disable this check , just
use a catch-all regex `.*`.


kleimkuhler (approved):

cpretzer (approved):
🚢

PR #1197, opened 2021-09-29
Add upgrade notes for 2.11

mateiidavid:
Added the upgrade notes and breaking changes.

### DIFF
--- a/linkerd.io/content/2.11/tasks/upgrade.md
+++ b/linkerd.io/content/2.11/tasks/upgrade.md
@@ -195,7 +195,7 @@ channel in the [Linkerd slack](https://slack.linkerd.io/).
 There are two breaking changes in the 2.11.0 release: pods in `ingress` no
 longer support non-HTTP traffic to meshed workloads; and the proxy no longer
 forwards traffic to ports that are bound only to localhost. Additionally, users
-of the multi-cluster extension will need to re-link their cluster after 
+of the multi-cluster extension will need to re-link their cluster after
 upgrading.
 
 ### Control plane changes
@@ -217,14 +217,10 @@ annotation on your ingress controller pod. In many cases, ingress mode is no
 longer necessary. Before upgrading, it may be worth revisiting [how to use
 ingress](../using-ingress/) with Linkerd.
 
-Second, the proxy will no longer forward traffic to ports bound on localhost.
-Previously, it was possible to have the proxy forward traffic to a service
-bound only on the loopback interface (such as `localhost:8080`). Typically,
-when a service listens only on localhost, it should not be accessible from the
-outside world. In order to prevent ports from being accidentally exposed to
-other pods in the cluster, the proxy will no longer forward to localhost, it
-will instead use the original address. This change should be reflected in your
-service configuration.
+Second, the proxy will no longer forward traffic to ports only bound on
+localhost, such as `127.0.0.1:8080`. Services that want to receive traffic from
+other pods should now be bound to a public interface (e.g `0.0.0.0:8080`). This
+change prevents ports from being accidentally exposed outside of the pod.
 
 ### Multicluster
 
--- a/linkerd.io/content/2.11/tasks/upgrade.md
+++ b/linkerd.io/content/2.11/tasks/upgrade.md
@@ -192,15 +192,18 @@ channel in the [Linkerd slack](https://slack.linkerd.io/).
 
 ## Upgrade notice: stable-2.11.0
 
-When upgrading, it is worth noting that the control plane
-has been changed to have a smaller footprint. The `controller` pod has been
-removed; consequently, all configuration options that previously applied to it
-are no longer valid (e.g `publicAPIResources` and all of its nested fields).
+There are two breaking changes in the 2.11.0 release: pods in `ingress` no
+longer support non-HTTP traffic to meshed workloads; and the proxy no longer
+forwards traffic to ports that are bound only to localhost. Additionally, users
+of the multi-cluster extension will need to re-link their cluster after 
+upgrading.
 
-Moreover, the destination pod will now have an additional `policy` container
-that runs the new policy controller. New healthchecks have been added to the
-CLI to help troubleshoot the state of Linkerd. There are some additional
-breaking changes in the 2.11.0 release that may affect you.
+### Control plane changes
+
+The `controller` pod has been removed from the control plane. All configuration
+options that previously applied to it are no longer valid (e.g
+`publicAPIResources` and all of its nested fields). Additionally, the
+destination pod has a new `policy` container that runs the policy controller.
 
 ### Routing breaking changes
 
--- a/linkerd.io/content/2.11/tasks/upgrade.md
+++ b/linkerd.io/content/2.11/tasks/upgrade.md
@@ -252,9 +252,7 @@ be aware of when upgrading from `stable-2.10.x`:
  list. Ports `4444`, `6379` and `9300` corresponding to Galera, Redis and
  ElasticSearch respectively (all server speak first protocols) have been added
  to the default opaque ports list. The default ignore inbound ports list has
- also been changed to include ports `4567` and `4568`. Note that when
- upgrading, any customization done to `proxy.opaquePorts` or
- `proxyInit.ignoreInboundPorts` will be overridden.
+ also been changed to include ports `4567` and `4568`.
 
 ## Upgrade notice: stable-2.10.0
 
--- a/linkerd.io/content/2.11/tasks/upgrade.md
+++ b/linkerd.io/content/2.11/tasks/upgrade.md
@@ -13,6 +13,7 @@ Before starting, read through the version-specific upgrade notices below, which
 may contain important information you need to be aware of before commencing
 with the upgrade process:
 
+- [Upgrade notice: stable-2.11.0](#upgrade-notice-stable-2-11-0)
 - [Upgrade notice: stable-2.10.0](#upgrade-notice-stable-2-10-0)
 - [Upgrade notice: stable-2.9.4](#upgrade-notice-stable-2-9-4)
 - [Upgrade notice: stable-2.9.3](#upgrade-notice-stable-2-9-3)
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+service configuration.
+
+### Multicluster
+
+In the multicluster extension, the nginx gateway deployment image has been
+replaced with a pause container. If you are currently running Linkerd 2.10.x
+together with the multicluster extension, in addition to upgrading the control
+plane and extension itself, you will also need to re-link all clusters. The
+gateway probes now target a port on the proxy instead of the now removed nginx
+container.
```
alpeb, line 234: Can you clarify that the upgrade won't result in downtime, but the link would need to be re-generated for `linkerd mc gateways` to properly report the status of the target clusters?

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+
+- `PodSecurityPolicy` (PSP) resources are no longer installed by default as a
+ result of their deprecation in Kubernetes v1.21 and above. The control plane
+ and core extensions will now be shipped without PSPs; they can be enabled
+ through a new install option `enablePSP: true`.
+- `tcp_connection_duration_ms` metric has been removed.
+- Opaque ports changes: `443` is no longer included in the default opaque ports
+ list. Ports `4444`, `6379` and `9300` corresponding to Galera, Redis and
+ ElasticSearch respectively (all server speak first protocols) have been added
+ to the default opaque ports list.
```
alpeb, line 255: Note that these changes will overwrite any customization done to `proxy.opaquePorts`.
Also `proxyInit.ignoreOutboundPorts` received a new default that will overwrite custom values.

alpeb (commented):
Looks great!
I just added a couple comments

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+version comes with fixes, improvements and new features, most notably support
+for inbound policies. When upgrading, it is worth noting that the control plane
+has been changed to have a smaller footprint. The `controller` pod has been
+removed; consequently, all configuration options that previously applied to it
+are no longer valid (e.g `publicAPIResources` and all of its nested fields).
+
+Moreover, the destination pod will now have an additional `policy` container
+that runs the new policy controller. New healthchecks have been added to the
+CLI to help troubleshoot the state of Linkerd. If you are upgrading from
+Linkerd 2.10.0, 2.10.1, or 2.10.2, there are some additional breaking changes
```
wmorgan, line 205: Since we only officially allow upgrades from these releases, I think the "If ... " clause here can be removed.

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
 version. If you have any questions, feel free to raise them at the #linkerd2
 channel in the [Linkerd slack](https://slack.linkerd.io/).
 
+## Upgrade notice: stable-2.11.0
+
+The upgrade procedure for `stable-2.11.0` is the same as before. The new
```
wmorgan, line 195: I would remove the first two sentences and, if it's possible to do so concisely, suggest a way to prune the controller configuration options.

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+Linkerd 2.10.0, 2.10.1, or 2.10.2, there are some additional breaking changes
+in the 2.11.0 release that may affect you.
+
+### Routing breaking changes
+
+There are two breaking changes to be aware of when it comes to how traffic is
+routed.
+
+First, when the proxy runs in ingress mode (`config.linkerd.io/inject:
+ingress`), non-HTTP outbound traffic is no longer supported. To get around
```
wmorgan, line 214: I don't think "outbound" is widely understood so I would say "non-HTTP traffic to meshed pods" instead. You might also point out that in many cases, ingress mode is no longer necessary and it's worth revisiting the [ingress docs](https://linkerd.io/2.10/tasks/using-ingress/) before upgrading.

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+this, you will need to use the `config.linkerd.io/skip-outbound-ports`
+annotation on your ingress controller pod.
+
+Second, the proxy will no longer forward traffic to ports bound on localhost.
+Previously, it was possible to have the proxy forward traffic to a service
+bound only on the loopback interface (such as `localhost:8080`). Typically,
+when a service listens only on localhost, it should not be accessible from the
+outside world. In order to prevent ports from being accidentally exposed to
+other pods in the cluster, the proxy will no longer forward to localhost, it
+will instead use the original address. This change should be reflected in your
```
wmorgan, line 224: This paragraph is a little repetitive.

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,70 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+service configuration.
+
+### Multicluster
+
+In the multicluster extension, the nginx gateway deployment image has been
+replaced with a pause container. If you are currently running Linkerd 2.10.x
```
wmorgan, line 230: Is there a one-sentence description of the motivation for this breaking change?

wmorgan (commented):
Thanks for tackling this!

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Yup, updated.

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Will do a bit of work, it felt repetitive to me too as I was writing.

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: No way that I know of to prune the config options but I'll have a look, just to be sure. Removed the first two lines.

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Done. Let me know if it looks ok

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Good point, updated.

linkerd.io/content/2.11/tasks/upgrade.md
alpeb: LGTM :+1:

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,73 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+
+When upgrading, it is worth noting that the control plane
+has been changed to have a smaller footprint. The `controller` pod has been
+removed; consequently, all configuration options that previously applied to it
+are no longer valid (e.g `publicAPIResources` and all of its nested fields).
+
+Moreover, the destination pod will now have an additional `policy` container
+that runs the new policy controller. New healthchecks have been added to the
+CLI to help troubleshoot the state of Linkerd. There are some additional
+breaking changes in the 2.11.0 release that may affect you.
```
wmorgan, line 203:
```suggestion
There are two breaking changes in the 2.11.0 release: pods in `ingress` no
longer support non-HTTP traffic to meshed workloads; and the proxy no longer
forwards traffic to ports that are bound only to localhost. Additionally, users
of the multi-cluster extension will need to re-link their cluster after 
upgrading.

### Control plane changes

The `controller` pod has been removed from the control plane. All configuration
options that previously applied to it are no longer valid (e.g
`publicAPIResources` and all of its nested fields). Additionally, the
destination pod has a new `policy` container that runs the policy controller.
```

linkerd.io/content/2.11/tasks/upgrade.md
```diff
@@ -189,6 +190,73 @@ Congratulation! You have successfully upgraded your Linkerd to the newer
...
+ result of their deprecation in Kubernetes v1.21 and above. The control plane
+ and core extensions will now be shipped without PSPs; they can be enabled
+ through a new install option `enablePSP: true`.
+- `tcp_connection_duration_ms` metric has been removed.
+- Opaque ports changes: `443` is no longer included in the default opaque ports
+ list. Ports `4444`, `6379` and `9300` corresponding to Galera, Redis and
+ ElasticSearch respectively (all server speak first protocols) have been added
+ to the default opaque ports list. The default ignore inbound ports list has
+ also been changed to include ports `4567` and `4568`. Note that when
+ upgrading, any customization done to `proxy.opaquePorts` or
```
wmorgan, line 257: I assume this is a Helm thing? If so, let's make that explicit. It sounds really scary otherwise.

linkerd.io/content/2.11/tasks/upgrade.md
wmorgan: Just trying to put the important stuff at the top.

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Okay, I've re-worded this. I wanted to make it more concise and keep it simple, there might be room to simplify further. Included examples of what it means for a port to be bound on localhost, and what it means for a service to be bound on a public interface. This should now: inform what the change is, what needs to change, and why the change was necessary.

Let me know if there's anything we can do to improve this or make it more clear.

linkerd.io/content/2.11/tasks/upgrade.md
mateiidavid: Yeah, it is related to Helm. After talking it through with @alpeb we decided to remove this note. By default the values will not be overridden. We also document how to handle values for upgrades. Users should have a clear idea on what to do.

PR #1200, opened 2021-09-29
Simplify emojivoto policy example

olix0r:
The example emojivoto policy was intially written to exercise a variety
of policy functionality, but isn't actually how we'd write policies for
emojivoto. Furthermore, because it uses an identity glob, it cannot work
in clusters with custom domains.

In order to make the example more representative of how policy would
actually be applied, this change unifies the emoji and voting
authorizations to a single authorization, `internal-grpc`, that selects
both the voting and emoji server by selector.


cpretzer (approved):
🚢

PR #1204, opened 2021-09-30
Add `theme-color` headers

olix0r:
The `<meta name="theme-color" content="..." />` pragma is used by
browsers to configure how the window should be decorated.  See
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta/name/theme-color

This change adds this directive to our meta.html template with a
hardcoded value that matches our header's background color. It seems
like we should be able to obtain this value from a template expression,
but I'm not sure how to do that.


olix0r:
image

PR #1208, opened 2021-10-01
clean up CNI docs

wmorgan:
Some people have been confused by the fact that they still need their own CNI plugin. Make that clear, and do some other cleanup.


cpretzer (approved):
🚢 LGTM

alpeb (commented):
LGTM, but to be super-precise, the actual capabilities are `CAP_NET_ADMIN` _and_ `CAP_NET_RAW`

PR #1209, opened 2021-10-05
Fix bring your own Jaeger instructions for 2.11

terrycain:
`collector.jaegerAddr` no longer works to configure the OpenCensus collector in 2.11 so this updates the documentation to patch the collector config.


linkerd.io/content/2.11/tasks/distributed-tracing.md
```diff
@@ -156,10 +156,51 @@ If you have an existing Jaeger installation, you can configure the OpenCensus
...
+      pipelines:
+        traces:
+          receivers: [otlp,opencensus,zipkin,jaeger]
+          processors: [batch]
+          exporters: [jaeger]
+EOF
+linkerd jaeger install --values ./jaeger-linkerd.yaml | kubectl apply -f -
 ```
 
+The YAML file is merged with the Helm values.yaml [here](https://github.com/linkerd/linkerd2/blob/stable-2.11.0/jaeger/charts/linkerd-jaeger/values.yaml)
```
cpretzer, line 201: What do you think of adding a note about the `jaeger.endpoint` entry to indicate that it should be set using the values appropriate for the environment?

In `values.yaml` the value is `jaeger.${POD_NAMESPACE}:14250` because the field is dynamically populated

cpretzer (commented):
thanks for great PR @terrycain I left a comment for you

PR #1216, opened 2021-10-17
fix up emojivoto policies to work with default:deny

wmorgan:
Add admin port policy bits, plus lots of comments, and reorder things a bit.

This works locally for me with a default-deny namespace annotation on emojivoto, but I would appreciate someone validating that.


wmorgan:
I agree to the DCO for all the commits on this PR :)

PR #1237, opened 2021-11-03
Correct checksum on installation

Bakrog:
The program shasum generates output in the following
format: `hash filename` and current scripts was
considering the filename as the hash.

Changed the regex to strip the filename and keep the hash.

Fixes linkerd/linkerd2#7202


Bakrog:
@alpeb hello, I did with my real email. It’s complaining because the commit headers was done using my GitHub no reply. Do so need to change the headers also?

alpeb (approved):
Ok I think it should be fine now. Thanks again!

PR #1242, opened 2021-11-22
A few versioning improvements before branching off the 2.12 docs

alpeb:
- Added new site param `latest_linkerd2_stable_version`
- Used that in `layouts/partials/docs.html` instead of having to hard-code the version
- Created shortcode `latest-linkerd2-stable-version` using that setting
- Used shortcode in `design-principles/_index.md` and `tasks/distributed-tracing.md` instead of hard-coded version
- Removed hard-coded version in `gsoc.html` and instead take advantage of alias url
- Replaced versioned url with relative url in `reference/iptables.md`


cpretzer (approved):
Great improvements @alpeb

PR #1264, opened 2021-12-23
Fix multicluster tutorial and kustomize resources

alpeb:
Fixes linkerd/linkerd2#7504

Kustomize breaking changes and other upgrades made that tutorial obsolete. Fixed with:

- Removed namespaces from resources and patch, and instead use `kubectl -n test apply`
- Having vars in the nginx' location targets implies declaring a dns resolver. Also extended FQDN for services is required.
- The nginx container requires running as root to allow using apk to install cURL

Also improved the tutorial itself:

- Added instructions for installing viz
- Removed not-required line-feeds in `echo` statements
- http://localhost:8080/east requires cURL and doesn't work in the browser because of javascript reasons
- Added missing directives in cleanup steps

### DIFF
--- a/linkerd.io/content/2.11/tasks/multicluster.md
+++ b/linkerd.io/content/2.11/tasks/multicluster.md
@@ -12,8 +12,8 @@ between services that live on different clusters.
 
 At a high level, you will:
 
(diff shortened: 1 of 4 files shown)


olix0r (approved):
lgtm

PR #1268, opened 2021-12-30
various docs updates

wmorgan:


linkerd.io/content/2.11/features/protocol-detection.md
```diff
@@ -31,30 +31,31 @@ likely running into a protocol detection timeout. This section will help you
...
 to destinations inside the cluster do not need further configuration.
-Linkerd's default list of opaque ports in the 2.11 release is 25 (SMTP), 587
-(SMTP), 3306 (MySQL), 4444 (Galera), 5432 (Postgres), 6379 (Redis), 9300
-(ElasticSearch), and 11211 (Memcache). Note that this may change in future
-releases.
 
-The following table contains common protocols that may require configuration.
+Linkerd's default list of opaque ports in the 2.11 release is **25** (SMTP),
+**587** (SMTP), **3306** (MySQL), **4444** (Galera), **5432** (Postgres),
+**6379** (Redis), **9300** (ElasticSearch), and **11211** (Memcache).
```
alpeb, line 55: Isn't this a little redundant with the table that follows? How about saying something like linkerd defaults to opaque all those ports in that table, unless otherwise noted?

alpeb (approved):

linkerd.io/content/2.11/features/protocol-detection.md
wmorgan: The intent is that the table contains all the "weird" protocols we discover (e.g. they sometimes arise in Slack) that need additional handling, and the list is what specifically is opaque'd by default. So the two are not necessarily the same.

PR #1276, opened 2022-01-12
chore: update lint and check command

mikutas:


cpretzer (commented):
Thanks for submitting this @mikutas 

Does it make sense to add the  `rm -rf tmp` to `make lint` or `make check`? If we want to rebuild the files before either of those commands, then maybe we should add the command then we could remove the files first. 

If not, then this PR looks good

mikutas:
tried to show example...

# Example

The `branch_a` has no problem so we expect `make check` always reports no error.

The `branch_b` contains a wrong file. If we are on `branch_b`, `make check` fails (as expected). 
`branch_a` doesn't contain the wrong file.

When we run `make check` on `branch_b` and move to `branch_a`, `tmp` still contains the wrong file (because git ignored).
Then we run `make check` on `branch_a`, it fails despite `branch_a` has no problem (because `tmp-sites` overwrites only files belonging to `branch_a`).

cpretzer (approved):
thanks @mikutas I appreciate the PR and the explanation/example

PR #1277, opened 2022-01-12
Require TLS when using curl

olix0r:
Our usage of `curl` could be vulnerable to protocol downgrade attacks. This
change updates most of our usage of curl--especially when fetching executable
resources--with the following command-line flags:

* `--proto '=https'` forces use of HTTPS. This ensures that dropping `https://`
  from a URL will cause the command to fail instead of reverting to use
  unsecured HTTP.
* `--tlsv1.2` disables the use of older TLS versions.
* `-f|--fail` ensures that curl does not output anything to stdout when a
  non-2xx response is received.
* `-S|--show-error` causes errors to be printed to stderr (when `-s|--silent`
  is used).

Related to linkerd/linkerd2#7593

Signed-off-by: Oliver Gould <ver@buoyant.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,14 +10,14 @@ RUN apt-get update \
   && export CLOUD_SDK_REPO="cloud-sdk-$(lsb_release -c -s)" \
   && echo "deb https://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" \
     | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list \
-  && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg \
+  && curl --proto '=https' --tlsv1.3 -sSfL https://packages.cloud.google.com/apt/doc/apt-key.gpg \
     | apt-key add - \
   && apt-get update \
   && apt-get install -y --no-install-recommends google-cloud-sdk \
   && wget https://github.com/gohugoio/hugo/releases/download/v0.61.0/hugo_extended_0.61.0_Linux-64bit.deb \
   && dpkg -i hugo*.deb \
   && rm hugo*.deb \
-  && curl https://htmltest.wjdp.uk | bash \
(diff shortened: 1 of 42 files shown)


alpeb (commented):
Looks that you mostly targeted 2.10 docs and not 2.11?

cpretzer (commented):
On macOS Big Sur, these commands fail because the default version of LibreSSL doesn't have TLS 1.3 support built in.

olix0r:
We should update this to be `--tls1.2`; and we should make sure the 2.11 docs are updated... I think I missed them via merge or something

hawkw (approved):

olix0r:
@cpretzer does this look okay now?

cpretzer (approved):
LGTM 🚢

klingerf:
Apologies for the late review here. The updated `curl` command from this branch fails for me locally:

```
$ curl --proto '=https' --tlsv1.2 -sSfL buoyant.cloud/install
curl: (1) Protocol "http" not supported or disabled in libcurl
```

```
$ curl --version
curl 7.77.0 (x86_64-apple-darwin21.0) libcurl/7.77.0 (SecureTransport) LibreSSL/2.8.3 zlib/1.2.11 nghttp2/1.42.0
Release-Date: 2021-05-26
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL UnixSockets
```

It's likely I could install a compatible version of curl via homebrew, but I wanted to mention it since I suspect other osx users may run into similar issues with the default curl install.

PR #1286, opened 2022-01-25
Adding explicit warning about why nginx ingress annotation is needed

juandiegopalomino:
And why folks might not want it.


cpretzer (commented):
@juandiegopalomino thanks for this PR. The docs include a mention of session stickiness in the section about the different [ingress modes](https://github.com/linkerd/website/blob/main/linkerd.io/content/2.11/tasks/using-ingress.md?plain=1#L528), do you think that this documentation covers the detail that you want to add?

juandiegopalomino:
@cpretzer nope because this caveat was not clear to me, and I released an upgrade with these changes that broke sticky sessions even though I was confident of the change because the nginx docs as they were were clear, straightforward, and did not mention this. 

Finding warning and potential issues like this should not be a scavenger hunt, they should be made loud and clear, even at the cost of repetition.

juandiegopalomino:
As a user, it took a lot of digging to understand exactly what was going on, why we needed this, and what was the sacrifice. With this change I hope that no other user will find it difficult.

mateiidavid (commented):
Hey @juandiegopalomino 

> Finding warning and potential issues like this should not be a scavenger hunt, they should be made loud and clear, even at the cost of repetition.

Sorry you feel this way. I personally think the instructions here aren't necessarily hard to find, there's a whole section at the bottom that details how ingress controllers are supposed to work conceptually (wrt routing), and how they integrate with Linkerd. 

Regardless, I know how all of this information can be a bit confusing, there's still some work to do if we want to make this simpler and more accurate. If you think the docs aren't particularly clear here, then by all means, we should fix that. Thank you for taking the initiative!

My only comment is that we should use the `{{ <note> }}` syntax. While we're here though, perhaps there's some better way to organise this? Maybe it would make more sense to put the note at the top instead of having it for every ingress controller?

wdyt @cpretzer

cpretzer:
> Hey @juandiegopalomino
> 
> > Finding warning and potential issues like this should not be a scavenger hunt, they should be made loud and clear, even at the cost of repetition.
> 
> Sorry you feel this way. I personally think the instructions here aren't necessarily hard to find, there's a whole section at the bottom that details how ingress controllers are supposed to work conceptually (wrt routing), and how they integrate with Linkerd.
> 
> Regardless, I know how all of this information can be a bit confusing, there's still some work to do if we want to make this simpler and more accurate. If you think the docs aren't particularly clear here, then by all means, we should fix that. Thank you for taking the initiative!
> 
> My only comment is that we should use the `{{ <note> }}` syntax. While we're here though, perhaps there's some better way to organise this? Maybe it would make more sense to put the note at the top instead of having it for every ingress controller?
> 
> wdyt @cpretzer

I'm in agreement that perhaps this can be addressed through reorganizing the docs, rather than duplicating the information on the page. Let's keep it DRY! 😄 

@juandiegopalomino can you suggest some reorganization of the content that would make the information more accessible?

juandiegopalomino:
Sorry, but I got other things to focus on now-- because my team and I now know of this issue and integrated it in our product, I don't need this update. I wished to do my part in helping the community, so I created this PR to help others experiencing the same issue. You and your colleagues have different opinions on the best way to do this update, and also don't seem sure that this update is needed. I'm sure that your stances are well-founded on maintaining this doc, but having said that debating what is the proper form is tedious, and not something I wish to get into. I therefore feel satisfied by simply raising the concern and offering a solution as I have done above. Merge, update, or ignore as you please-- it will be all good with me. Thank you for your time, and I promise to raise other PRs as I find concerns with the docs!

cpretzer:
@juandiegopalomino I hear you and we appreciate you taking the time to submit the PR and have this dialog with us. Apologies if it was tedious. 

I'll ask another colleague to review, just to get another perspective. 

Thanks again!

alpeb:
I tried merging this one up but it requires merging main, resolving conflicts and add the changes into the 2.12 and 2-edge docs as well, and upstream doesn't have write permissions. So closing for now.

PR #1291, opened 2022-01-28
Add info about multicluster policy for extensions

klingerf:
The default policy that's included with Linkerd's multicluster extension restricts access to the multicluster service mirror components to only the prometheus instance that's installed by the linkerd viz extension. Update the multicluster task doc with information about configuring policy for other extensions that require it, namely Buoyant Cloud.

Signed-off-by: Kevin Ingelman <ki@buoyant.io>


cpretzer (approved):
LGTM 🚢

PR #1307, opened 2022-02-24
Update using-ingress.md - Nginx tips

qingvincentyin:
The write-up for Nginx controller is inadequate. It took me a long time to troubleshoot my problem even though I thought I followed your docs. While your docs weren't wrong, it will be a lot better to add the few paragraphs I'm proposing.


linkerd.io/content/2.11/tasks/using-ingress.md
```diff
@@ -100,6 +100,29 @@ spec:
         number: 80
 ```
 
+The `namespace` containing the ingress controller should NOT be annotated with `linkerd.io/inject: enabled`. Rather, annotate the `kind: Deployment` (`.spec.template.metadata.annotations`) of the nginx.
+
+If using [this Helm chart](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx), then the `values.yaml` should look like this:
```
cpretzer, line 105: Can we make this more generic? Something like "Be sure that the values.yaml file for your ingress helm chart includes the YAML below"

qingvincentyin: That segment of `values.yaml` only applies to that particular Helm chart. If someone invents a new chart to deploy the Nginx ingress, it'll likely be a different segment -- different parameter name, different nesting structure, etc.

Also, the chart I referred to is the _community_ version of Nginx ingress controller. There's a competing NGINX controller, too. See:

-  https://www.nginx.com/blog/guide-to-choosing-ingress-controller-part-4-nginx-ingress-controller-options/

The latter one will have its own Helm chart which I am not familiar with.

The point being, the content of `values.yaml` is highly dependent on which chart we choose. The one I chose is a pretty popular chart.

Anyway, the sample segment of `values.yaml` is only an example. The paragraph before that explains the underlying reason and goal of the configuration.

qingvincentyin: OK. I've rephrased my paragraphs in the spirit of what you suggested. It's now clear that the discussion is only for that particular Helm chart (a popular chart.)

cpretzer (approved):
thanks @qingvincentyin 🚢

PR #1308, opened 2022-02-28
Update canary-release.md: add prereq of SMI-extension

qingvincentyin:
It's not obvious in the docs that we need to install the Linkerd SMI-extension before following this tutorial. This PR adds that prereq.


cpretzer (approved):
Thanks, @qingvincentyin !

PR #1312, opened 2022-03-01
Update run.linkerd.io to point to stable 2.11

kleimkuhler:
This applies some auto formatting as well.

Signed-off-by: Kevin Leimkuhler <kleimkuhler@icloud.com>

### DIFF
--- a/run.linkerd.io/public/index.html
+++ b/run.linkerd.io/public/index.html
@@ -1,1 +1,12 @@
-<!DOCTYPE html><html><head><title>https://linkerd.io/</title><link rel="canonical" href="https://linkerd.io/2.10/getting-started/" /><meta name="robots" content="noindex"><meta http-equiv="content-type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="0; url=https://linkerd.io/2.10/getting-started/" /></head></html>
+<!DOCTYPE html>
+<html>
+
+<head>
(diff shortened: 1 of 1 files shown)


cpretzer:
@kleimkuhler I wonder if the html was intentionally unformatted to make the response smaller?

run.linkerd.io/public/index.html
```diff
@@ -1 +1,12 @@
-<!DOCTYPE html><html><head><title>https://linkerd.io/</title><link rel="canonical" href="https://linkerd.io/2.10/getting-started/" /><meta name="robots" content="noindex"><meta http-equiv="content-type" content="text/html; charset=utf-8" /><meta http-equiv="refresh" content="0; url=https://linkerd.io/2.10/getting-started/" /></head></html>
+<!DOCTYPE html>
+<html>
+
+<head>
+    <title>https://linkerd.io/</title>
+    <link rel="canonical" href="https://linkerd.io/2.11/getting-started/" />
```
alpeb, line 6: Actually let's make this just `2`, which cloudflare automatically redirects to whatever the latest version is
```suggestion
    <link rel="canonical" href="https://linkerd.io/2/getting-started/" />
```

kleimkuhler:
@cpretzer I don't think that's really a concern here. The difference in response size is only going to be the `\n`s being added which is going to be ~24 bytes (12 lines).

PR #1313, opened 2022-03-02
Remove Ambassador instructions from multi-cluster doc

olix0r:
We've had a number of users get confused by the Ambassador instructions in the
multi-cluster docs. This change removes these instructions.

Signed-off-by: Oliver Gould <ver@buoyant.io>


cpretzer (approved):
LGTM 🚢

cpretzer:
This is probably in the 2.12 branch, too

We do get questions about using ingress for MC gateways, so I think some docs might be useful. Or perhaps a tutorial/guide in a different section

alpeb:
> This is probably in the 2.12 branch, too

No need to worry about the 2.12 branch. From time to time I'm bringing that one in sync :-)

PR #1314, opened 2022-03-03
Wrong Service Name

cawoodm:
The service is called `prometheus` and it's in the `linkerd-viz` namespace.


cawoodm:
Not only, followed your getting started guide and the service is named "prometheus".

kleimkuhler (approved):

PR #1353, opened 2022-05-13
Add env variable to publish workflow

mateiidavid:
The publish step seems to fail when trying to build the website. Inside the build script, a `CURRENT_SHA` env variable is set to the latest git commit sha. When building the website, however, the contents are moved to a temp directory that does not appear to be a git repository. Because of this, attempting to do `git rev-parse` to set the value of `CURRENT_SHA` fails.

This change removes the unused `CURRENT_SHA` variable to allow the site to be built.

The `CURRENT_SHA` variable appears to be unused

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -26,7 +26,5 @@ jobs:
         gcloud auth activate-service-account --key-file=.gcp.json
 
     - name: Publish
-      env:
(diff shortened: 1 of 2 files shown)


olix0r (approved):

adleong (approved):
Would be worth updating the description of this PR to say that this variable appears to be unused.

PR #1362, opened 2022-05-31
Skip old versions in sitemap

alpeb:
Docs under "/2.10" are getting indexed in Google instead of those under "/2.11". To fix that, this refactors the `sitemap.xml` file in two ways:

- Added a condition to skip any versioned page that doesn't point to the latest version
- Replace version in URL with "2", which Cloudflare permanently redirects to the latest version


wmorgan (approved):
Resulting sitemap.xml looks good to me.

cpretzer (approved):
LGTM: https://deploy-preview-1362--linkerdio.netlify.app/sitemap.xml

PR #1367, opened 2022-06-08
Fix small typo: desireable --> desirable

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@fujitsu.com>


cpretzer (approved):
good catch, thanks @hs0210

PR #1379, opened 2022-07-01
Add NetworkPolicy notes to protocol detection

mateiidavid:
NetworkPolicies that restrict ingress access should target the proxy's
inbound port (4143) when the application's exposed port is marked as
opaque. The outbound proxy will re-write the target port based on the
opaque protocol hint set by the destination controller.


cpretzer (approved):
LGTM 🚢

PR #1382, opened 2022-07-13
Add Calico CNI configuration information, fix note formatting

dtaskai:
Added configuration information for the Calico CNI to the `Cluster Configuration` page based on the information provided by @mateiidavid on Slack and also added this to the note on the `Getting Started` page (the one mentioning additional configuration steps in case of a GKE private cluster) and fixed the note formatting to match the other notes featured on the website.


linkerd.io/content/2.11/reference/cluster-configuration.md
```diff
@@ -75,3 +75,16 @@ Finally, verify that the firewall is created:
...
 gcloud compute firewall-rules describe gke-to-linkerd-control-plane
 ```
+
+## Calico CNI
+
+For Calico to be able to enforce network policies it needs to get the Pod's IP
+address from the Kubernetes API Server, but the `postStart` hooks introduced
+in 2.11 interfere with this. Calico's network policy won't let traffic through
+until it has acquired the Pod's IP address, and the Pod can't progress until it
+has network connectivity.
```
mateiidavid, line 85: I think this is fine, but again, I'd like to be a bit more generic and articulate more on what the problem is. Here's how I would do it, feel free to add your own spin to it if you want:

```suggestion
Linkerd uses a `postStart` lifecycle hook for all control plane components, and all injected workloads by default. The hook will poll proxy readiness through [linkerd-await(https://github.com/linkerd/linkerd-await) and block the main container from starting until the proxy is ready to handle traffic. By default, the hook will time-out in 2 minutes.

CNI plugins that are responsible for setting up and enforcing `NetworkPolicy` resources can interfere with the lifecycle hook's execution. While lifecycle hooks are running, the container will not reach a `Running` state. Some CNI plugin implementations acquire the Pod's IP address only after all containers have reached a running state, and the kubelet has updated the Pod's status through the API Server. Without access to the Pod's IP, the CNI plugins will not operate correctly. This in turn will block the proxy from being set-up, since it does not have the necessary network connectivity.
```

mateiidavid: It's a bit more verbose but I think it'll save people time from having to look-up the actual details since we're being more thorough in our explanation.

linkerd.io/content/2.11/reference/cluster-configuration.md
```diff
@@ -75,3 +75,16 @@ Finally, verify that the firewall is created:
...
+
+Until this behavior is fixed, manually removing the `postStart` hooks in case
+of the control plane components and annotating either the workloads and/or
+namespaces with `config.linkerd.io/proxy-await: disabled` in the case of data
+plane components solves this issue.
```
mateiidavid, line 90: This is great! I'd only change the start; we'd ideally want this piece of documentation to hold and not have to change it between releases. As of `stable-2.11.x`, we can only remove the hook manually on the control plane components. I wouldn't specify "Until the behavior is fixed", since imo that'd require changes to the docs and us keeping up with the state of other CNI implementations. Not to mention, people may run outdated versions that do not contain the fix.

```suggestion
As a workaround, users can manually remove the `postStart` lifecycle hook from control plane components. For injected workloads, users may opt out of the lifecycle hook through the root-level `await: false` option, or alternatively, behavior can be overridden at a workload or namespace level through the annotation `config.linkerd.io/proxy-await: disabled`.  Removing the hook will allow containers to start asynchronously, unblocking network connectivity once the CNI plugin receives the pod's IP.
```

Something like this :)

mateiidavid (commented):
Sorry, it took me a while to get to this :) I left a few comments. I think the text is fine as is, but it would good to tidy it up a little bit and make the problem & solution a bit more explicit, without pointing at just a singular CNI plugin implementation (there may be more plugins that receive the pod IP the same way as Calico).

It also looks as if a merge has gone slightly wrong here, we have a few files that might not be associated with this change. It might be a good idea to do a soft reset and only check-in the relevant files. Usually merges will add the whole git log, a `git pull` to merge might make it a bit easier since we will only have one merge commit (that will be squashed when the PR is merged anyway).

dtaskai:
I've added your suggestions by hand to avoid the need to fix the length in another commit, git still seems to be messed up, it wouldn't let me push the changes without pulling first, let me know on Slack (to avoid cluttering the PR) if something has to be done with it! :smile:

mateiidavid:
Hey @dtaskai, seems that there is a conflict which I think might be unrelated to your changes. I'd try to do the following to fix the git files:

```
$ git reset --soft ORIG_HEAD # or commit prior to you making these changes
# 
# Your commit history in 'git log' should be reset back to when you started the work
# your changes will still be there, however, they just have to be staged again.
$ git add <file>
$ git commit -s
# 
# After you commit, you'll have a bunch of unrelated files there
# it's safe to discard them, likely created by the merge
$ git checkout . 
$ git pull origin main
# 
# You should now be on the latest main and only have your change
# you might need to push with --force, since the commit history is
# different between your local history and the branch history
$ git push origin <branch> -f 
```

That should hopefully fix your commit history. Might take a bit of trial and error unfortunately, but it'll be easier to fix your branch and easier for us to review the change :)

mateiidavid (approved):
Perfect! Thanks for the work @dtaskai!

PR #1386, opened 2022-07-20
Fix Grafana Helm commands

siggy:
The commands at https://linkerd.io/2.11/tasks/grafana/#install-grafana
had a line break in the wrong place.

Fixup the Grafana Helm commands.


alpeb (approved):
Thanks @siggy !

PR #1388, opened 2022-07-21
Fix broken relative links

jeremychase:
The four links below were broken. It appears we attempted to convert them to relative links and missed removing the domain.

Signed-off-by: Jeremy Chase <jeremy.chase@gmail.com>


kflynn (approved):
Augh. Yes.

PR #1413, opened 2022-08-22
move data plane proxy metrics check to `linkerd-viz` section

hawkw:
This check is run as part of `linkerd viz check --proxy`, _not_ 
`linkerd check --proxy`, but it's currently in the section "The
'linkerd-data-plane' checks". I've moved it to the section "The
'linkerd-viz' checks, which I think is where it should be.

### DIFF
--- a/linkerd.io/content/2.12/tasks/troubleshooting.md
+++ b/linkerd.io/content/2.12/tasks/troubleshooting.md
@@ -1091,31 +1091,6 @@ For more information on `linkerd inject`, see
 [Step 5: Install the demo app](../../getting-started/#step-5-install-the-demo-app)
 in our [Getting Started](../../getting-started/) guide.
 
-### √ data plane proxy metrics are present in Prometheus {#l5d-data-plane-prom}
-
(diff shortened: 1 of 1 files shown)


hawkw:
Some additional questions:
- I noticed we don't have a separate section for the `linkerd viz check --proxy` checks. Should we add one, or does it make sense to put these in the same section as the `linkerd viz check` checks that run without `--proxy`?
- The hint anchor for this check doesn't include `l5d-viz` (it's `#l5d-data-plane-prom` instead of `#l5d-viz-data-plane-prom`). We may want to change this, but that would require changing the CLI as well, so I didn't do that in this PR.

hawkw:
The [`markdownlint` failure](https://github.com/linkerd/website/runs/7956688514?check_suite_focus=true) is unrelated; it looks like the same check [is failing on the `alpeb/2.12` branch](https://github.com/linkerd/website/runs/7911493490?check_suite_focus=true).

adleong (approved):

PR #1415, opened 2022-08-22
upgrade: remove `$` from bash snippets without output

hawkw:
Markdownlint [suggests][1] that bash snippets showing commands to be run
only include a `$` prefix before the command when the snippet also
includes output from that command. This is why the CI build for the
`alpeb/2.12` branch [is currently failing][2]. 

To fix CI, this branch disables that rule (MD014) in `tasks/upgrade.md`.
As far as I can tell, Markdownlint doesn't support some kind of config
file or other way to disable a rule for the whole repo...

[1]: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md014---dollar-signs-used-before-commands-without-showing-output
[2]: https://github.com/linkerd/website/runs/7911493490?check_suite_focus=true

### DIFF
--- a/linkerd.io/content/2.12/tasks/upgrade.md
+++ b/linkerd.io/content/2.12/tasks/upgrade.md
@@ -214,18 +214,18 @@ need to add the following annotations and label to the CRD so that the
 `linkerd-smi` chart can adopt it:
 
 ```bash
-kubectl annotate --overwrite crd/trafficsplits.split.smi-spec.io \
+$ kubectl annotate --overwrite crd/trafficsplits.split.smi-spec.io \
   meta.helm.sh/release-name=linkerd-smi \
   meta.helm.sh/release-namespace=linkerd-smi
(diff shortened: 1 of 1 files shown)


adleong (approved):
I think that the showing a prompt makes these easier to read, especially in snippits that include multiple commands to emphasize that multiple commands are being shown (as opposed to a single multi-line command).  I'd lean slightly toward disabling this particular markdown-lint rule, but this is not a hill I'd die on.

hawkw:
> I'd lean slightly toward disabling this particular markdown-lint rule, but this is not a hill I'd die on.

we could also do that --- the reason this is a lint is because, when no prompt is shown, the commands can be copied and pasted without having to remove the `$`. 

alternatively, we could consider using `:;` as a "prompt" character, since a command prefixed with `:;` is not a syntax error if copied and pasted without removing the prompt character.

hawkw:
@adleong I've changed this PR to disable the rule, instead.

Issue #1403, opened 2022-08-22
CLI doc updates for `linkerd authz` (needs to note that it can show AuthorizationPolicy now)


jeremychase:

adleong:
Fixed by https://github.com/linkerd/website/pull/1420

PR #1423, opened 2022-08-23
Update restricting access docs with new linkerd viz authz output

adleong:
Fixes https://github.com/linkerd/website/issues/1403


linkerd.io/content/2.12/tasks/restricting-access.md
```diff
@@ -162,72 +166,17 @@ following logic when deciding whether to allow a request:
...
-    name: admin
-  client:
-    networks:
-    - cidr: 10.244.0.1/32
-    unauthenticated: true
-```
+If a port does not have a Server defined, Linkerd will automatically use a
+default Server which allows readiness and liveness probes. However, if you
+create a Server resource for a port which handles probes, you will need to
+explicitly create an authorization to allow those probe requests.
```
hawkw, line 179: maybe worth linking to the doc on creating per-route policies added in #1426 here ?

hawkw (approved):

PR #1429, opened 2022-08-23
Add feature page for iptables modes

### LINKED ISSUE #1411: Feature description for `iptables-nft`

mateiidavid:
Since the introduction of 'nft' mode, linkerd-init can be run in different
modes. This PR adds a feature page to detail how the two modes differ, and how
config can be overridden. Additionally, we mention the two modes in the
architecture page.

Closes #1411

### DIFF
--- a/linkerd.io/content/2.12/features/nft.md
+++ b/linkerd.io/content/2.12/features/nft.md
@@ -5,10 +5,10 @@ description = "Linkerd's init container can run in two separate modes, nft or le
 
 To transparently route TCP traffic through the proxy, without any awareness
 from the application, Linkerd will configure a set of [firewall
-rules](../../reference/iptables) in each injected pod. Configuration can be
+rules](../../reference/iptables/) in each injected pod. Configuration can be
 done either through an [init
 Container](../../reference/architecture/#linkerd-init-container) or through a
-[CNI plugin](../cni)
+[CNI plugin](../cni/)
 
 Linkerd's init container can be run in two separate modes: `legacy` or `nft`.
 The difference between the two modes is what variant of iptables they will use
--- a/linkerd.io/content/2.12/features/nft.md
+++ b/linkerd.io/content/2.12/features/nft.md
@@ -1,5 +1,5 @@
 +++
-title = "Proxy Init Modes"
+title = "Proxy Init Iptables Modes"
 description = "Linkerd's init container can run in two separate modes, nft or legacy."
 +++
 
@@ -7,11 +7,11 @@ To transparently route TCP traffic through the proxy, without any awareness
 from the application, Linkerd will configure a set of [firewall
 rules](../../reference/iptables/) in each injected pod. Configuration can be
 done either through an [init
-Container](../../reference/architecture/#linkerd-init-container) or through a
+container](../../reference/architecture/#linkerd-init-container) or through a
 [CNI plugin](../cni/)
 
 Linkerd's init container can be run in two separate modes: `legacy` or `nft`.
-The difference between the two modes is what variant of iptables they will use
+The difference between the two modes is what variant of `iptables` they will use
 to configure firewall rules.
 
 ## Details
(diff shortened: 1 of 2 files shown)


linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
++++
+title = "Proxy Init Modes"
+description = "Linkerd's init container can run in two separate modes, nft or legacy."
```
hawkw, line 3: neither the title or the description say anything about iptables, maybe we should mention the word "iptables" here? it seems like users looking for information on this might be looking for the word "iptables" in the title?
```suggestion
title = "Proxy Init Iptables Modes"
description = "Linkerd's init container can run in two separate modes, nft or legacy."
```

linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
...
++++
+
+To transparently route TCP traffic through the proxy, without any awareness
+from the application, Linkerd will configure a set of [firewall
+rules](../../reference/iptables/) in each injected pod. Configuration can be
+done either through an [init
+Container](../../reference/architecture/#linkerd-init-container) or through a
```
hawkw, line 10: not sure if "container" should be capitalized here?
```suggestion
container](../../reference/architecture/#linkerd-init-container) or through a
```

linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
...
+[CNI plugin](../cni/)
+
+Linkerd's init container can be run in two separate modes: `legacy` or `nft`.
+The difference between the two modes is what variant of iptables they will use
```
hawkw, line 14: we never really say what iptables _is_, do we want to briefly define/introduce the term? or link to upstream linux docs?

linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
...
+installation. Once configured, all injected workloads (including the control
+plane) will use the same mode in the init container. Both modes will use the
+`iptables` utility to configure firewall rules; the main difference between the
+two, is which binary they will call into:
+
+1. `legacy` mode will call into `iptables-legacy` for firewall configuration.
+   This is the default mode that `linkerd-init` runs in, and is supported by
+   most operating systems and distributions.
+2. `nft` mode will call into `iptables-nft`, which uses the newer `nf_tables`
+   kernel API. The `nftables` utilities are used by newer operating systems to
```
hawkw, line 29: nit: we say `nf_tables` and then we say `nftables`, is that correct? should we use one name or the other, or are both used interchangeably?

linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
...
+   configure firewalls by default.
+
```
hawkw, line 31: it might be worth linking to some kind of reference material when we first mention `iptables-legacy` and `iptables-nft`. maybe the man pages?
```suggestion
1. `legacy` mode will call into [`iptables-legacy`] for firewall configuration.
   This is the default mode that `linkerd-init` runs in, and is supported by
   most operating systems and distributions.
2. `nft` mode will call into `iptables-nft`, which uses the newer `nf_tables`
   kernel API. The [`nftables`] utilities are used by newer operating systems to
   configure firewalls by default.

[`iptables-legacy`]: https://manpages.debian.org/bullseye/iptables/iptables-legacy.8.en.html
[`iptables-nft`]: https://manpages.debian.org/bullseye/iptables/xtables-nft.8.en.html
```

another good external doc to reference for more details might be might be https://developers.redhat.com/blog/2020/08/18/iptables-the-two-variants-and-their-relationship-with-nftables

linkerd.io/content/2.12/features/nft.md
```diff
@@ -0,0 +1,60 @@
...
+## Installation
+
+The mode for `linkerd-init` can be overridden through the configuration option
+`proxyInit.mode=iptables|nft`. The configuration option can be used for both
+Helm and CLI installations (or upgrades). For example, the following line will
+install Linkerd and set the init container mode to `nft`:
+
+```bash
+linkerd install --set "proxyInit.mode=nft" | kubectl apply -f -
+```
```
hawkw, line 60: should we also provide an example of a Helm install without using the Linkerd CLI?

linkerd.io/content/2.12/reference/architecture.md
```diff
@@ -102,4 +102,5 @@ The `linkerd-init` container is added to each meshed pod as a Kubernetes [init
 container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/)
 that runs before any other containers are started. It [uses
 iptables](../iptables/) to route all TCP traffic to and from the pod through
-the proxy.
+the proxy. Linkerd's init container can be run in [different
+modes](../../features/nft/)
```
hawkw, line 106: just saying "different modes" here is a little bit vague, maybe we should say something like this:
```suggestion
the proxy. Linkerd's init container can be run in [different
modes](../../features/nft/) which determine what iptables variant is used.
```

hawkw (approved):
this looks good overall, although i have some small suggestions.

i'm not totally sure if "Features" is the right place for this document to go; is this "Features" or "Reference" content? i could be convinced either way...

also, i wonder if the "iptables" doc in the "Reference" docs should also link to this document? https://github.com/linkerd/website/blob/1e24fd64e7d3f43ceb87d6e46da4fe5fd11d9c5a/linkerd.io/content/2.12/reference/iptables.md

adleong (approved):

linkerd.io/content/2.12/features/nft.md
mateiidavid: we can use them interchangeably tbh, the utilities are generally called `nftables` (like the actual binaries), but the modules appear as `nf_tables`. Whichever one would pick is good as long as we stay consistent.

PR #1438, opened 2022-08-23
add access logging docs

### LINKED ISSUE #1417: Access Logging

hawkw:
This branch adds a "Tasks" page documenting the proxy's access
logging functionality, added in 2.12. I've also added a note to the
docs on configuring the proxy's log level that states that access 
logging is a distinct feature that is configured separately.

Closes #1417


hawkw:
I put this in "Tasks", but alternatively, it could go in "Features". I noticed that the announcement blog post currently has an (incorrect) link to `/features/access-logging`. Does anyone have an opinion on what section this documentation ought to go in? The existing logging docs are all in "Tasks", but I'm happy to move it.

Either way, I'll be sure to update the release blog post with the final URL.

adleong (approved):

linkerd.io/content/2.12/tasks/enabling-access-logging.md
```diff
@@ -0,0 +1,64 @@
...
+
+Linkerd proxies can be configured to generate an HTTP access log that records
+all HTTP requests that transit the proxy.
+
+The `config.linkerd.io/access-log` annotation is used to enable proxy HTTP
+access logging. Adding this annotation to a namespace or workload configures the
+proxy injector to set an environment variable in the proxy container that
+configures access logging.
+
+Enabling HTTP access logging has a performance cost, and may introduce
```
jeremychase, line 18: @hawkw Is there any way we can quantify this or give an indication how severe the performance cost is?

hawkw: we haven't done particularly exhaustive testing, but see https://github.com/linkerd/linkerd2-proxy/pull/1319#issuecomment-1016956871

hawkw: personally i would want to err on the side of not making very precise claims, because the answer to any performance question is always going to be "it depends"

wmorgan:
The boundary is a little fuzzy but I would prefer this to be a feature rather than a task.

linkerd.io/content/2.12/tasks/enabling-access-logging.md
hawkw: we could put the "5%" number here, but it's going to depend somewhat on the nature of the traffic --- a proxy handling a bunch of short HTTP/1.1 requests will write to the access log a lot more frequently than a proxy handling long-running streaming HTTP/2 requests, because the access log is written when a request ends. Personally, I think the best thing to do is to say "increased CPU time and/or tail latencies"; if you care about the _precise_ performance impact on your workload, you should test it for your specific workload.

jeremychase: Understandable perspective. How about this wording:

> HTTP access logging is disabled by default because it consumes additional CPU time compared to proxies without access logging enabled. Under load you are likely to experience high tail latencies and reduced throughput. The severity of this performance penalty depends on the nature of your traffic and may be acceptable in your environment.

hawkw: i went with something similar, with some slight changes (mostly to remove second-person language, which we don't use elsewhere in this doc)

wmorgan (approved):
🥳

linkerd.io/content/2.12/tasks/enabling-access-logging.md
jeremychase: That was a test.

PR #1439, opened 2022-08-23
regen cli-2-12.yaml One Last Time (TM)

hawkw:


adleong (changes requested):
I think you generated this on an older version of Linkerd or a branch that was missing some of the latest changes.  Note in the diff a bunch of regressions.

adleong (approved):

PR #1445, opened 2022-08-24
remove needless `cat <<EOF`s in 2.12 docs

hawkw:
In several places, the "Tasks" documentation shows applying a Kubernetes
manifest by `cat`ing a heredoc and piping it into `kubectl apply -f -`.
In most cases, the `cat <<EOF | kubectl apply -f -` is unneeded, and can
be simplified to just `kubectl apply -f - <<EOF`.

I didn't touch the cases where the yaml is first piped into 
`linkerd inject -` and *then* into `kubectl apply`, as the cat is necessary as
the first stage in those pipelines (I think).


kflynn (approved):
Thanks! I think that looks cleaner. 🙂

PR #1449, opened 2022-08-24
update 2.12 docs to use `linkerd install --crds` before `install`

### LINKED ISSUE #1454: Update the install instructions in the generating certificates tutorial
The installation instructions in https://linkerd.io/2.12/tasks/generate-certificates/#passing-the-certificates-to-linkerd are out of date for both CLI and Helm methods.

The CLI instructions should be updated to include the `install --crds` step.

The Helm instructions should be updated to reference the `linkerd-crds` chart.

### LINKED ISSUE #1457: Multicluster install tutorial needs crds step
The multicluster tutorial needs to be updated to include the `install --crds` install step

https://linkerd.io/2.12/tasks/multicluster/#install-linkerd-and-linkerd-viz

(Also, the install anchor link in the TOC is broken)

hawkw:
The Linkerd CRDs must be installed before installing the control plane.
Most of the "Tasks" docs that require Linkerd to be installed just
reference the "Install" task, which demonstrates the use of
`install --crds`. However, some of the docs demonstrate alternative
configurations or installation methods, and actually show running the
`install` command. Many of those docs don't indicate that
`linkerd install --crds` must be run first.

This branch updates the 2.12 "Tasks" docs to show the use of
`linkerd install --crds` prior to running a `linkerd install` command. I
think I got all of them? Similarly, when showing Helm installs, we now
install the `linkerd-crds` chart before installing other Helm charts.

I didn't add `install --crds` in a couple of places where the install
command is not piped directly into `kubectl apply`. For example, in the
private docker registry documentation, we show piping `linkerd install`
into `grep` to get the docker image versions, and in that case, we are
not actually installing Linkerd into a cluster in that command example,
so the installation won't actually fail.

Fixes #1457
Fixes #1454


adleong (approved):

kleimkuhler (approved):

kflynn (approved):
Looks great, hit it! 🙂

PR #1476, opened 2022-08-31
Document different options for shutdown

### LINKED ISSUE #1402: document shutdown behavior, including wait-before-exit-seconds, shutdown-grace-period (2.12), linkerd-await, etc
The general problem of pod shutdown is getting complicated. We now have two config variables as of 2.12 and a helper binary. This should all be documented with a little explainer of how Kubernetes shutdown policy works in the first place.

@mateiidavid had some good commentary in the Slack:

> wait-before-exit-seconds will just delay the handling of the TERM signal for the proxy, when the signal comes from the kubelet. If a container doesn't handle the TERM signal within the graceful shutdown period, then all containers will be killed. The proxy when it receives a TERM signal will generally wait until all connections have finished (yet will not establish new connections). When connections fail to close (for whatever reason), in 2.12 you'll be able to set the graceful period in the proxy through [annotations](https://github.com/linkerd/linkerd2/pull/8923) so regardless of how many open connections it has, it will shutdown.  However, these solutions only work when your container gets killed due to:
* preemption/descheduling;
* liveness/readiness probe failing;
* resource contention;
* kubectl delete, etc.

That might be a good starting point.

mateiidavid:
Fixes #1402

### DIFF
--- a/linkerd.io/content/2.12/tasks/graceful-shutdown.md
+++ b/linkerd.io/content/2.12/tasks/graceful-shutdown.md
@@ -15,11 +15,78 @@ job resources.
 
 ## Graceful shutdown in Kubernetes
 
+[pod-lifetime]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-lifetime
+[pod-termination]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
+[pod-forced]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-forced
+[hook]: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks
+
+Pods are ephemeral in nature, and may be [killed due to a number of different
+reasons][pod-lifetime], such as:
+
+* Being scheduled on a node that fails (in which case the pod will be deleted).
+* A lack of resources on the node where the pod is scheduled (in which case the
+  pod is evicted).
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2.12/tasks/graceful-shutdown.md
```diff
@@ -13,6 +13,81 @@ calls after the proxy has received the TERM signal, those network calls will
...
+
+Linkerd offers a few options to configure pods and containers to gracefully shutdown.
+
+* `--wait-before-seconds`: can be used as an install value (either through the
+  CLI or through Helm), or alternatively, through a [configuration
+  annotation](../../reference/proxy-configuration.md). This will add a
+  `preStop` hook to the proxy container to delay its handling of the TERM
+  signal. This will only work when the conditions described above are satisfied
+  (i.e container runtime sends the TERM signal)
+* `config.linkerd.io/shutdown-grace-period`: is an annotation that can be used
```
wmorgan, line 70: Very nice writeup. The only thing I don't see is a use case for the `shutdown-grace-period` annotation... when is that necessary?

wmorgan (approved):

linkerd.io/content/2.12/tasks/graceful-shutdown.md
mateiidavid: Great point, we generally want to use `shutdown-grace-period` when we have long running connections that prevent the proxy from shutting down. Added a short sentence to describe that under the bullet point.

PR #1483, opened 2022-09-02
Install redirects for old canonicals:

kflynn:
2.12/reference/cli/repair -> 2.12/reference/cli because the repair subcommand is no more
2.12/tasks/securing-your-cluster -> 2.12/tasks/securing-linkerd-tap because the content seems to have been renamed
2.12/tasks/upgrading-2.10-ports-and-protocols -> 2.12/features/protocol-detection because that page was about upgrading _to_ 2.10 and getting protocol detection
2.11/tasks/upgrading-2.10-ports-and-protocols -> 2.11/features/protocol-detection, same as far 2.12.


wmorgan (approved):
Makes sense to me!

Issue #1496, opened 2022-09-16
2.12 Helm installation documentation mentions obsolete installNamespace variable


svend:
The 2.12 [Helm install documentation](https://github.com/linkerd/website/blob/264f7f31be757eb5eca9baed936dfe8c50c25ce1/linkerd.io/content/2.12/tasks/install-helm.md?plain=1#L100-L113) mention the `installNamespace` variable, but that was [removed prior to 2.12](https://github.com/linkerd/linkerd2/pull/7492).

``` markdown
## Customizing the namespace

To install Linkerd to a different namespace, you can override the Helm
`Namespace` variable.

By default, the chart creates the control plane namespace with the
`config.linkerd.io/admission-webhooks: disabled` label. This is required for the
control plane to work correctly. This means that the chart won't work with
Helm's `--namespace` option.  If you're relying on a separate tool to create the
control plane namespace, make sure that:

1. The namespace is labeled with `config.linkerd.io/admission-webhooks: disabled`
1. The `installNamespace` is set to `false`
1. The `namespace` variable is overridden with the name of your namespace
```

The [namespace resource](https://github.com/linkerd/linkerd2/blob/ee75526ba7caddd3df2ddd4d3e3393c599715d4b/charts/linkerd-control-plane/templates/namespace.yaml#L1) in the chart contains the following conditional: `{{- if eq .Release.Service "CLI" -}}`. According to the [Helm documentation](https://helm.sh/docs/chart_template_guide/builtin_objects/), `.Release.Service` is always `Helm` when using `helm`.

kleimkuhler:
Closed by #1497

PR #1497, opened 2022-09-19
Remove obsolete instructions for namespace customization

alpeb:
As of 2.12.0, when installing via Helm, helm itself should create the namespace (or some other step in the CD pipeline), and we no longer require any special metadata on it.


alpeb:
Closes #1496

wmorgan (approved):
Ty!

PR #1500, opened 2022-09-22
Fix typo

### LINKED ISSUE #1477: Update version on protocol-detection page
On this page: https://linkerd.io/2.12/features/protocol-detection/

The following text should be updated: "As of the 2.11 release" to "As of the 2.12 release"

mikutas:
just Fixes #1477


alpeb:
Checking the values.yaml file for stable-2.11.4, I see those entries were already there, so the doc is correct in stating that as of 2.11 the list is such.

mikutas:
@alpeb I guess users reading this page expect to know default opaque ports as of 2.12 (even if the ports are not changed since 2.11).

The description `"As of the 2.11 release"` on 2.12 docs may possibly cause doubts.

But, if there is no need to update, I will close this.

alpeb (approved):
Ok you're right, it's less confusing if that matches the docs version :-)

PR #1504, opened 2022-09-25
Update control-plane version in snippet.

### LINKED ISSUE #1503: Using GitOps with Linkerd with Argo CD
I have a documentation bug to report.  I was loosely following the tutorial here for setting up LInkerD with ArgoCD - https://linkerd.io/2.12/tasks/gitops

But when I got to the part of installing the linkerd-control-plane - the example provided was attempting to install version 2.12 from the chart - but it doesn't look like the control plane follows that versioning scheme.  The latest control plane I was able to find was 1.9.3.

I guess this is either an issue with the control plane repo - or the docs :)

dtaskai:
Fixes #1503, but should probably be merged the same time as https://github.com/linkerd/linkerd-examples/pull/272 for the versions to match on both sides.

Signed-off-by: Dominik Táskai <dominik.taskai@leannet.eu>

### DIFF
--- a/linkerd.io/content/2.12/tasks/gitops.md
+++ b/linkerd.io/content/2.12/tasks/gitops.md
@@ -399,7 +399,7 @@ Ensure that the multi-line string is indented correctly. E.g.,
   source:
     chart: linkerd-control-plane
     repoURL: https://helm.linkerd.io/stable
-    targetRevision: 1.9.3
+    targetRevision: 1.9.0
     helm:
       parameters:
       - name: identityTrustAnchorsPEM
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2.12/tasks/gitops.md
```diff
@@ -399,7 +399,7 @@ Ensure that the multi-line string is indented correctly. E.g.,
   source:
     chart: linkerd-control-plane
     repoURL: https://helm.linkerd.io/stable
-    targetRevision: 2.12.0
+    targetRevision: 1.9.3
```
alpeb, line 402: Should be 1.9.0 to point to 2.12.0
```suggestion
    targetRevision: 1.9.0
```

Issue #1512, opened 2022-10-17
Document workaround for initContainers with CNI network plugin


bartwitkowski:
Please add 3 workarounds to CNI plugin documentation here [https://linkerd.io/2.12/features/cni/](https://linkerd.io/2.12/features/cni/)

1. Add annotation config.linkerd.io/skip-outbound-ports
2. Add annotation config.linkerd.io/skip-subnets (waiting for [#9565](https://github.com/linkerd/linkerd2/issues/9565) to be solved)
3. Add configuration initContainer UID and set it to 2102 (same as used in linkerd-proxy sidecar)

More info and discussion in [#9625](https://github.com/linkerd/linkerd2/issues/9625) issue.

bartwitkowski:
Documentation updated.

PR #1513, opened 2022-10-19
Document UID workaround #9625

jeremychase:
This PR adds a section titled "Allowing initContainer networking" to the "CNI Plugin" Feature documentation.

Intended to address #1512 and [linkerd2 issue #9625](https://github.com/linkerd/linkerd2/issues/9625)

Signed-off-by: Jeremy Chase <jeremy.chase@gmail.com>

image


stevej:
Question: where does user id 2101 come from? Can it be different in other clusters?

linkerd.io/content/2.12/features/cni.md
```diff
@@ -94,6 +94,29 @@ The most important flags are:
    `journalctl -t kubelet`. The string `linkerd-cni:` can be used as a search to
    find the plugin log output.
 
+### Allowing initContainer networking
+
+When using the Linkerd CNI plugin the required `iptables` rules are in effect
+before the pod is scheduled. Also, the `linkerd-proxy` is not started until
+after all `initContainers` have completed. This means no `initContainer` will
+have network access because its packets will be caught by `iptables` and the
+`linkerd-proxy` will not yet be availble.
```
mateiidavid, line 103: nit:
```suggestion
`linkerd-proxy` will not yet be available.
```

linkerd.io/content/2.12/features/cni.md
```diff
@@ -94,6 +94,29 @@ The most important flags are:
...
+
+It is possible to bypass these `iptables` rules by running the `initContainer`
+as the UID `2102`. Processes run as this UID are skipped by `iptables` and allow
```
mateiidavid, line 106: TIOLI:
```suggestion
as the UID of the proxy (by default `2102`). Processes run as this UID are skipped by `iptables` and allow
```

We allow people to configure the UID if they really want to. This is useful in certain distributions where each namespace gets a range of UIDs to be used by pods. I think we can save ourselves some trouble if we say "skip proxy uid which by default is 2102".

Alternatively, we could on L120 include a note:

```
{{< note >}}
By default the proxy is set to run with the UID 2102. If you are overriding this value in your manifests, the overridden UID needs to be used instead.
{{< /note >}}

mateiidavid (approved):
One tiny typo and a suggestion. Giving the shipit signal since everything else looks perf.

🚢 📦

stevej (approved):
LGTM

kleimkuhler (approved):

PR #1523, opened 2022-11-16
Document TLS version and cipher suite used for mTLS connections

dan-ri:
Document the TLS version and cipher suite used for mTLS connections, as this information may be required for security audit purposes in some deployment environments.

### DIFF
--- a/linkerd.io/content/2.12/features/automatic-mtls.md
+++ b/linkerd.io/content/2.12/features/automatic-mtls.md
@@ -135,6 +135,15 @@ identity name. When the proxy connects to the destination, it initiates a TLS
 handshake and verifies that that the destination proxy's certificate is signed
 by the trust anchor and contains the expected identity.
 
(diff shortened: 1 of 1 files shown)


kleimkuhler (commented):
This looks good. It'd be nice to specify that these values are set [here](https://github.com/linkerd/linkerd2-proxy/blob/500dc3a1279a0be92a912136a4c54e213236b7a7/linkerd/meshtls/rustls/src/creds.rs#L119-L121), but I don't really want to put a link to code in the documentation incase files change in the future.

So aside from the nit that I left, this should be fine.

hawkw (approved):
This looks good to me! I agree that, while adding a reference to the source code could be nice, it's more effort to ensure that the docs are up to date.

I wonder if we should also document the kind of keys that are generated somewhere on this page?

kleimkuhler (approved):
Thanks @dan-ri!

PR #1547, opened 2022-12-21
Remove no-longer valid note about `automountServiceAccountToken`

alpeb:
Fixes linkerd/linkerd2#9970

Removed note stating that `automountServiceAccountToken` needs to be enabled in a pod in order to be injectable. That is no longer the case after `edge-21.11.1` after which we're adding Service Account Token Volume projections.


wmorgan:
@alpeb To confirm, this applies to all 2.12 releases?

alpeb:
> @alpeb To confirm, this applies to all 2.12 releases?

That's correct. It was added into 2.12.0

kleimkuhler (approved):

PR #1557, opened 2023-01-20
Update authorization policy documenation with wildcard support

tmablunar:
Documenation for AuthorizationPolicy was missing support for wildcards in `identities` as documented in the [CRD](https://github.com/linkerd/linkerd2/blob/904a9316af28d015779943dcbbdbc9d380a73a5d/charts/linkerd-crds/templates/policy/meshtls-authentication.yaml#L45) and in the [issue](https://github.com/linkerd/linkerd2/issues/7709) introducing targetRefs.


kleimkuhler (approved):
Thanks @tmablunar!

PR #1560, opened 2023-01-20
docs: only response total metrics have classifications

hawkw:
Currently, the documentation for proxy metrics states that the `status_code`, `grpc_status_code`, and `classification` labels are present on "all `response_*` metrics". This is incorrect. The `response_latency_ms` metric does not include `classification` and `grpc_status_code` labels, since it is recorded when the first byte of a response body is received, and the response is not classified (and the gRPC status is not known) until the response body _ends_. The incorrect docs have caused confusion for users (see linkerd/linkerd2#9737).

This branch updates the proxy metrics documentation to distinguish between the labels that are present on all `response_*` metrics, and the labels that are present only on `response_total` metrics. In addition, I added a note explaining _why_ some labels are only present on the `response_total` metrics and not on `response_latency_ms`.

Fixes linkerd/linkerd2#9737

### DIFF
--- a/linkerd.io/content/2.12/reference/proxy-metrics.md
+++ b/linkerd.io/content/2.12/reference/proxy-metrics.md
@@ -72,13 +72,30 @@ Each of these metrics has the following labels:
 
 The following labels are only applicable on `response_*` metrics.
 
+* `status_code`: The HTTP status code of the response.
+
+#### Response Total Labels
+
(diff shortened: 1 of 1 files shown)


kleimkuhler (approved):

kflynn (approved):
This looks much better – many thanks!! 🙂

PR #1562, opened 2023-01-25
Add Linkerd Day EU 2023 entry to the Community menu

kflynn:
In theory, the external URL should cause the link to open in a
separate window.


wmorgan:
What about pointing to the blog post instead? That Kubecon site is just a generic CFP for all zero day events. I'd rather give context about the event plus pointers for sponsors, attendees, etc

wmorgan (approved):

PR #1565, opened 2023-02-03
Adds a note about installing grafana on 2.12 and its initContainer

danibaeyens:
Coming from this conversation:
https://linkerd.slack.com/archives/C89RTCWJF/p1675336124844379?thread_ts=1673519822.380309&cid=C89RTCWJF

I think this note can be useful for people who wants to install grafana on newer versions.


danibaeyens:
`make lint` do wonders 😇

I added the Signed-off... why is not accepting it?

danibaeyens:
Now looks better :)

kleimkuhler:
Thanks for the clarifying the docs on this! It's worth noting that this issue only occurs if you are using the CNI plugin. Without the CNI plugin, then the `linkerd-init` initContainer should be running after any other initContainers, which means iptables haven't changed yet.

With the CNI plugin, iptables have changed before any initContainers run which is why you'd run into this issue. So, it's probably worth saying in this note that if you are using the CNI plugin, then you will run into this issue and you should solve it with the ways you have listed.

linkerd.io/content/2.12/tasks/grafana.md
```diff
@@ -41,6 +41,17 @@ Cloud](https://grafana.com/products/cloud/). Those projects provide instructions
 on how to easily import the same charts published on
 <https://grafana.com/orgs/linkerd>.
 
+{{< note >}}
+Grafana's official Helm chart uses an initContainer to download Linkerd's
+configuration and dashboards. If you use the CNI plugin, when you add grafana's
+pod into the mesh, its initContainer run before the proxy is started, and the
```
kleimkuhler, line 47:
```suggestion
pod into the mesh its initContainer will run before the proxy is started and the
```

kleimkuhler (commented):
Thanks! Looks good; just small typo and maybe removing some commas.

kleimkuhler (approved):
Thank you so much @danibaeyens!

PR #1573, opened 2023-02-11
Fix pronunciation guide for "linkerd"

djanderson:
This commit suggests that the pronunciation of linkerd could be more clearly written as "linker-dee" instead of "linkerd-dee".


hawkw (approved):

wmorgan:
I think this is failing because a tweet elsewhere on the site was deleted. @kflynn can you take a look?

PR #1580, opened 2023-03-02
Add edge docs

jeremychase:
This PR adds edge documentation of Linkerd at `https://linkerd.io/2-edge/overview/`. The new documentation in `content/2-edge` has yet to be updated to match current edge behavior.

Notes:
* This PR does not add a link to the new edge docs in the version pulldown.
* When viewing `2-edge` documentation, the version pulldown will indicate the user is viewing `Linkerd [edge]`.
* The reason we are using `2-edge` rather than `edge` is because our website has already used the `https://linkerd.io/edge` URL and we do not want to break existing links.
* This PR adds a message on `2-edge` pages indicating that the documentation is not for the current stable version.

Screenshot of `2-edge` overview:

image


linkerd.io/layouts/partials/docs.html
```diff
@@ -44,6 +44,21 @@
               {{ end }}
             </div>
           </div>
+          {{ else if and (eq $docsVersion "2-edge") }}
```
kflynn, line 47: Surely we don't need the `and` here? I mean, I don't really care, just noticing. 😂

kflynn (approved):
This looks like a great idea, hit it! (I saw one place we might be able to delete an `and`, but I'm not gonna hold it for that.)

linkerd.io/layouts/partials/docs.html
jeremychase: Fixed. 🤣

PR #1587, opened 2023-03-15
Set `-n linkerd` in helm HA install examples

siggy:
The `helm install` commands with `values-ha.yaml` were not setting the `-n linkerd` flag, causing linkerd to be installed into the `default` namespace.

Modify the `helm install` examples to set `-n linkerd`.


alpeb (approved):
Oh great catch!

PR #1592, opened 2023-03-21
Add header-based routing tutorial

alpeb:
You can check the render [here](https://github.com/linkerd/website/blob/7cbb84c77df9ba7cbfd00e0d70614ff546063075/linkerd.io/content/2-edge/tasks/configuring-dynamic-request-routing.md)

### DIFF
--- a/linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
+++ b/linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
@@ -1,0 +1,140 @@
++++
+title = "Configuring Header-Based Routing"
+description = "Configuring HTTPRoute resources to perform header-based routing."
++++
+
+## Prerequisites
+
+To use this guide, you'll need to have Linkerd installed on your cluster. Follow
+the [Installing Linkerd Guide](../install/) if you haven't already done this
+(make sure you have at least linkerd stable-2.13.0 or edge-23.3.2).
+
+You also need to have the [Helm](https://helm.sh/docs/intro/quickstart/) CLI
+installed.
+
+## HTTPRoute for Header-Based Routing
+
+With header-based routing you can route HTTP traffic given the contents of
+request headers. This can be useful for performing load-balancing, A/B testing
+and many other strategies for traffic management.
+
+In this tutorial, we'll make use of the
+[podinfo](https://github.com/stefanprodan/podinfo) project to showcase
+header-based routing, by deploying in the cluster two backend and one frontend
+podinfo pods. Traffic will flow to just one backend, and then we'll switch
+traffic to the other one just by adding a header to the frontend requests.
+
+## Set Up
+
+First we create the `test` namespace, annotated by linkerd so all pods that get
+created there get injected with the linkerd proxy:
+
+``` bash
+kubectl create ns test --dry-run -o yaml \
+  | linkerd inject - \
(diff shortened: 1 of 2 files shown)


linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
```diff
@@ -0,0 +1,140 @@
...
+the [Installing Linkerd Guide](../install/) if you haven't already done this
+(make sure you have at least linkerd stable-2.13.0 or edge-23.3.2).
+
+You also need to have the [Helm](https://helm.sh/docs/intro/quickstart/) CLI
+installed.
+
+## HTTPRoute for Header-Based Routing
+
+With header-based routing you can route HTTP traffic given the contents of
+request headers. This can be useful for performing load-balancing, A/B testing
```
hawkw, line 18: how is header-based routing used for performing load balancing here?

linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
```diff
@@ -0,0 +1,140 @@
...
+    - matches:
+      - headers:
+        - name: "x-request-id"
+          value: "alternative"
+      backendRefs:
+        - name: "backend-b-podinfo"
+          port: 9898
+    - backendRefs:
+      - name: "backend-a-podinfo"
+        port: 9898
```
hawkw, line 105: it might be worth including at least a brief discussion of the structure of the HTTPRoute used here and how the route's rules actually work? In particular, it might be worth discussing the rules for how a rule is chosen?

linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
```diff
@@ -0,0 +1,140 @@
...
+  parentRefs:
+    - name: backend-a-podinfo
+      kind: Service
+      group: core
+      port: 9898
+  rules:
+    - matches:
+      - headers:
+        - name: "x-request-id"
+          value: "alternative"
```
hawkw, line 99: it could be worth mentioning somewhere that we can have multiple rules that match different values of the same header, and that header values can also be matched using regular expressions?

linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
alpeb: I was just thinking on rudimentary load balancing. Either manual or assisted by some controller not specified here. I can remove this if it's confusing or can be misinterpreted.

linkerd.io/content/2-edge/tasks/configuring-header-based-routing.md
alpeb: Agreed, I'll add more detail :+1:

prajithp13:
Can you please add an example for TrafficSplit as well? also, is it possible to have multiple values?

wmorgan:
@alpeb can we rename this feature from "header-based routing" to "dynamic request routing" in these docs please?

kflynn (approved):
This looks good from here.

alpeb:
> Can you please add an example for TrafficSplit as well? also, is it possible to have multiple values?

Yeah traffic split is also considered under the dynamic request routing umbrella. We'll provide instructions/examples to that as a followup :+1:

PR #1603, opened 2023-04-11
New directory for 2.13 docs

alpeb:
As instructed in the internal `linkerd-docs` procedure.


kflynn:
@alpeb Is this just `cp 2-edge 2.13`?

adleong (approved):

kflynn (approved):
Looks great, hit it! 🙂

alpeb:
> @alpeb Is this just `cp 2-edge 2.13`?

Unfortunately it's not just that, but we also require a bunch of other small changes in other places.

kflynn:
I should've phrased that better: yeah, there are metadata changes, but it seems that the content of the docs is what was on 2-edge.

alpeb:
> I should've phrased that better: yeah, there are metadata changes, but it seems that the content of the docs is what was on 2-edge.

Correct

PR #1609, opened 2023-04-14
Update fault injection docs to use HTTPRoute

adleong:
Instead of SMI TrafficSplit.


hawkw (approved):
nice! should we update the `2-edge` docs as well?

alpeb (approved):
LGTM :+1: 
The idea for the `2-edge` directory is to duplicate the current stable one, plus any changes that'll come in the next major version, so yeah, any change to stable docs should go into the edge docs as well.

PR #1613, opened 2023-04-24
add ServiceProfile/HTTPRoute compatibility warnings

hawkw:
Proxies will not honor configuration from both the `OutboundPolicies`
policy controller API and the `DestinationProfiles` destination
controller API for the same destination. Instead, one or the other is
used. This means that ServiceProfiles (which generate
`DestinationProfiles` configuration) and outbound HTTPRoutes and failure
accrual annotations (which generate `OutboundPolicies`) are mutually
incompatible, which isn't currently mentioned anywhere in the
documentation.

This branch adds warnings to the documentation that uses HTTPRoutes and
failure accrual annotations so that users are aware of this.


hawkw:
IIRC, @kflynn suggested removing the mention of "the `OutboundPolicies` API" until we actually have documentation defining that term. I'm going to do that before merging.

PR #1615, opened 2023-04-27
Add eBPF socket lvl lb details

mateiidavid:
Fixes https://github.com/linkerd/linkerd2/issues/10788

It's possible for CNI plugins to interfere with some features when resolving services directly at the socket level. In kube-proxy replacement mode, Cilium establishes client sockets that are connected directly to the service backend, thereby bypassing Linkerd's service discovery. As a consequence, dynamic request routing (and of course load balancing) will not work correctly.


linkerd.io/content/2.13/reference/cluster-configuration.md
```diff
@@ -76,6 +76,35 @@ Finally, verify that the firewall is created:
 gcloud compute firewall-rules describe gke-to-linkerd-control-plane
 ```
 
+## eBPF CNIs
+
+When the network is configured through a CNI plugin that replaces kube-proxy
+with an eBPF solution, certain Linkerd features may not work correctly. eBPF-based CNI plugins may assign sockets to service backends directly
```
wmorgan: Are these two lines extra?

linkerd.io/content/2.13/reference/cluster-configuration.md
```diff
@@ -76,6 +76,35 @@ Finally, verify that the firewall is created:
...
+When the cluster network is configured through eBPF-based CNI plugins, some
+Linkerd features may not work correctly. In general, certain CNI plugins may be
+configured to replace kube-proxy's routing functionality with an eBPF solution.
+If kube-proxy functionality is replaced, in-cluster service connections
+(East/West traffic) may have their sockets assigned to service backends
+directly, bypassing the intermediate hop to a `ClusterIP` service.
+
+In practice, this clashes with Linkerd's service discovery and can result in
+undefined behavior. Binding the socket directly to a backend during connection
+establishment (i.e bypassing the `ClusterIP`) means that Linkerd will forward
```
wmorgan:
```suggestion
establishment (i.e. bypassing the `ClusterIP`) means that Linkerd will forward
```

linkerd.io/content/2.13/reference/cluster-configuration.md
```diff
@@ -76,6 +76,35 @@ Finally, verify that the firewall is created:
 gcloud compute firewall-rules describe gke-to-linkerd-control-plane
 ```
 
+## eBPF CNIs
```
wmorgan, line 79: Why not just make this whole thing about Cilium? There's no theoretical set of other eBPF CNIs out there and even if they were it's not clear they would all behave this way.

mateiidavid: There's also Calico. We've had issues with their ebpf data plane _when talking to the API Server_. Their docs are a bit unclear on whether they do the same socket level load balancing for everything or just the API Server though. Your suggestion is great though, I'll just stick to Cilium here.

linkerd.io/content/2.13/reference/cluster-configuration.md
mateiidavid: Yes. Edited them out and simplified the problem a little bit (less mentions of sockets).

linkerd.io/content/2.13/reference/cluster-configuration.md
log1cb0mb: I can confirm that during my tests, I reproduced the same behaviour in my environments running Calico with eBPF datapath however from quick search, I could not find the equivalent config to cilium's `bpf-lb-sock-hostns-only` on calico side. Will have to dig deeper into their documentation or check with their dev if its even something configurable in current verions.

adleong (approved):

adleong (commented):
We should also add this content to the `2-edge` docs, and even potentially add it to previous versions as well.

kflynn (approved):
Hit it! 🙂

PR #1616, opened 2023-05-02
Add user logos: tradeshift and adidas

dered-dev:
**ADD** Tradeshift and Adidas logos in home page
**REMOVE** Bigcommerce  and HP logos in home page


kflynn (approved):
Hit it!! 🙂

PR #1581, opened 2023-03-02
corrects small typo in rotating tls certs docs

kevin-benton:
Very small typo fix - changes `whcih` to `which` in the Automatically Rotating Control Plan TLS Credentials page in the 2.12 docs.

PR #1622, opened 2023-05-08
fix typo which in auto rotate control plane tls

Siegfriedk:
Was fixed in 2.12 here https://github.com/linkerd/website/pull/1581 but not in 2.13 / edge.


hawkw (approved):
Thanks for the fix!

PR #1631, opened 2023-05-25
OSM migration blog post

kflynn:


wmorgan:
Nice. Is there a third migration strategy where you have both running side-by-side, and you flip one service at a time? Is that feasible?

kflynn:
@wmorgan I'm playing with that before writing about it -- since they're both sidecar meshes, I think you might need to make sure that the services are on disjoint Nodes, which is probably not a great answer. But I should know more shortly.

wmorgan (approved):

PR #1634, opened 2023-06-02
Adds notes about l5d-dst-override requirement

chicocvenancio:
2.13 changed behavior for ingress mode proxies,
failling requests without l5d-dst-override.

This add notes about this behaviour change to the docs

Fixes linkerd/linkerd2#10908


chicocvenancio:
Close since behavior has been identified as a bug.

wmorgan:
Thank you for this anyways, @chicocvenancio! I put a note about the bug in #1642

PR #1637, opened 2023-06-12
Drop forestry

kflynn:
False positives from this directory are annoying in searches.


JasonMorgan (approved):
LGTM

PR #1642, opened 2023-06-22
clarify ingress docs and add note about l5d-dst bug

wmorgan:


kflynn (approved):
Approving so that @wmorgan can merge if he thinks my commit is OK.

hawkw (approved):

PR #1643, opened 2023-06-23
Apparently a double slash in an image URL passes website lint but doesn't render correctly?? Also fix the alt text. Sigh.

kflynn:
It makes me tremendously sad that this doesn't look right on the website, but it passed lint _and_ rendered correctly in my local preview. 😢 


JasonMorgan (approved):
LGTM

PR #1655, opened 2023-08-11
Introduce pod-to-pod communication outline for edge docs

mateiidavid:
Still a work in progress


linkerd.io/content/2-edge/features/flat-network-multicluster.md
```diff
@@ -0,0 +1,45 @@
++++
+title = "Multi-cluster communication in flat networks"
+description = "Multi-cluster communication using pod-to-pod communication inshared network environments."
```
alpeb, line 3:
```suggestion
description = "Multi-cluster communication using pod-to-pod communication in shared network environments."
```

linkerd.io/content/2-edge/features/multicluster.md
alpeb, line 1: I think it'd be good to introduce both modes from the beginning, so users don't have to wait till way down the doc to discover about the p2p mode.

linkerd.io/content/2-edge/features/flat-network-multicluster.md
```diff
@@ -0,0 +1,91 @@
++++
+title = "Multi-cluster communication in flat networks"
```
wmorgan, line 2: Just looking at this from the outside, it feels like this separate feature doc shouldn't exist, and any additional content should be rolled into the existing multicluster feature doc.

wmorgan:
I've updated these docs to my semi-satisfaction, including moving the second doc to the reference section where I think it was intended to be. There is some repetition between feature & reference docs which is not ideal but is palatable.

There is a section at the end of the reference doc about namespace sameness that I left in, but I think this is more confusing than helpful and I'm not sure what the actual implications are for users. Keep your control plane namespaces the same? I would like to remove this section but perhaps I'm missing somethign.

wmorgan:
Also: there is still one <placeholder> reference that needs to be fixed once we have the URL.

alpeb (approved):

PR #1657, opened 2023-08-18
add top-level HTTPRoute features doc

hawkw:
Currently, the website doesn't really have a top-level summary of what
the HTTPRoute resource is and what it can do, similar to the
ServiceProfiles feature page. I thought this would be useful to add, as
it can be used for other documentation to link to, and will also be
useful as a place to collect all the documentation on things users can
do using HTTPRoutes. In addition, we can also add stuff about GAMMA
support and about the difference between `policy.linkerd.io` and
`gateway.networking.kubernetes.io` HTTPRoutes here, as well.

I've also updated some of the other documentation that mentions the
HTTPRoute resource to link to this page.


linkerd.io/content/2-edge/features/httproute.md
```diff
@@ -0,0 +1,62 @@
...
+  policy includes [dynamic request routing][dyn-routing], adding request
+  headers, modifying a request's path, and reliability features such as
+  [timeouts].
+- HTTPRoutes with a [Server] as their parent resource configure policy for
+  _inbound_ proxies in pods which recieve traffic to that [Server]. Inbound
+  HTTPRoutes are used to configure fine-grained [per-route authorization and
+  authentication policies][auth-policy].
+
+{{< warning >}}
+Outbound HTTPRoutes are **incompatible with ServiceProfiles**. If a
```
wmorgan, line 32: If this is a temporary situation, I would call it out as such and say that this will be fixed in the future. If this is a permanent situation, I would say that ServiceProfiles override HTTPRoute behavior rather than saying they're incompatible, which to me suggests that something will break.

linkerd.io/content/2-edge/features/server-policy.md
```diff
@@ -87,11 +87,11 @@ authentication rules to targets.
   unless an authentication rule is met
 
 - `ServerAuthorization`: an earlier form of policy that restricts access to
-  `Servers` only (i.e. not `HTTPRoutes`)
+  [`Server`s] only (i.e. not [`HTTPRoute`]s)
 
 The general pattern for Linkerd's dynamic, fine-grained policy is to define the
 traffic target that must be protected (via a combination of `Server` and
-`HTTPRoute` CRs); define the types of authentication that are required before
+[`HTTPRoute`] CRDs); define the types of authentication that are required before
```
wmorgan, line 94: was this change from CR to CRDs intentional?

linkerd.io/content/2-edge/features/server-policy.md
```diff
@@ -113,9 +113,9 @@ from Kubernetes, meaning that the pod would not be able to start. Thus, any
 default-deny setup must, in practice, still authorize these probes.
 
 In order to simplify default-deny setups, Linkerd automatically authorizes
-probes to pods. These default authorizations apply only when no `Server` is
-configured for a port, or when a `Server` is configured but no `HTTPRoutes` are
-configured for that `Server`. If any `HTTPRoute` matches the `Server`, these
+probes to pods. These default authorizations apply only when no [`Server`] is
+configured for a port, or when a [`Server`] is configured but no [`HTTPRoute`]s are
+configured for that [`Server`]. If any [`HTTPRoute`] matches the `Server`, these
```
wmorgan, line 118: I applaud the consistency in hyperlinking though I don't know if we've followed that pattern elsewhere in the site.

linkerd.io/content/2-edge/tasks/configuring-timeouts.md
```diff
@@ -4,17 +4,11 @@ description = "Configure Linkerd to automatically fail requests that take too lo
 +++
 
 To limit how long Linkerd will wait before failing an outgoing request to
-another service, you can configure timeouts. These work by adding a little bit
-of extra information to the [service profile](../../features/service-profiles/) for
+another service, you can configure timeouts. Timeouts can be specified using
+either [HTTPRoute] [service profile](../../features/service-profiles/) for
```
wmorgan, line 8:
```suggestion
either [HTTPRoute]s or [service profile](../../features/service-profiles/) for
```

wmorgan (approved):

linkerd.io/content/2-edge/features/server-policy.md
hawkw: i thought this was a typo, but I think it's not actually...i'll undo this change!

linkerd.io/content/2-edge/features/httproute.md
hawkw: i'm happy to reword this. i should mention, though, that the text of this warning is copied from other docs pages --- do you think it makes sense to also apply the same change to other instances of this warning box in the docs?

hawkw: @wmorgan what do you think of the wording in https://github.com/linkerd/website/pull/1657/commits/3f5b04edbd0662c9cfa677a91ee9182562336804 ? I'm happy to make the same change to other warnings if you like that language!

linkerd.io/content/2-edge/features/httproute.md
```diff
@@ -0,0 +1,64 @@
...
+  HTTPRoute.
+-->
+
+- [Configure dynamic request routing][dyn-routing] using an outbound HTTPRoute.
+- [Configure per-route authorization policy][auth-policy] using an inbound
+  HTTPRoute.
+- See the [reference documentation](../../reference/httproute/) for a complete
+  description of the HTTPRoute resource.
+
+[HTTPRotue resource]: https://gateway-api.sigs.k8s.io/api-types/httproute/
```
kflynn, line 58: s/HTTPRotue/HTTPRoute/

linkerd.io/content/2-edge/features/server-policy.md
```diff
@@ -87,11 +87,11 @@ authentication rules to targets.
   unless an authentication rule is met
 
 - `ServerAuthorization`: an earlier form of policy that restricts access to
-  `Servers` only (i.e. not `HTTPRoutes`)
+  [`Server`s] only (i.e. not [`HTTPRoute`]s)
```
kflynn, line 90: The "s" in [`Server`s] needs to move outside the brace to match your link definition.

linkerd.io/content/2-edge/features/server-policy.md
```diff
@@ -87,11 +87,11 @@ authentication rules to targets.
...
 
 The general pattern for Linkerd's dynamic, fine-grained policy is to define the
 traffic target that must be protected (via a combination of `Server` and
-`HTTPRoute` CRs); define the types of authentication that are required before
+[`HTTPRoute`] CRs); define the types of authentication that are required before
 access to that traffic is permitted (via `MeshTLSAuthentication` and
 `NetworkAuthentication`); and then define the policy that maps authentication to
 target (via an `AuthorizationPolicy`).
```
kflynn, line 97: Might this be better as a bulleted list?

> The general pattern ... is:
> - define the traffic target ...;
> - define the types of auth...; and finally
> - define the policy... .

Thoughts? (I know you weren't editing this for content really, just occurred to me while reading it.)

kflynn (changes requested):
I really like this! I've left some comments for minor things that I think need fixing, but really, they're minor, nice job. 🙂 

One final note: in some places you talk about e.g. "the `HTTPRoute` blah blah" and in other places it's "the HTTPRoute blah blah". We should probably pick one way and go with it. (IIRC the K8s standard is not to set type names in monospace, so that would be my vote, but it's not a terribly strong opinion...)

linkerd.io/content/2-edge/features/server-policy.md
hawkw: i think it might be better as a bulleted list, but i'd prefer not to make that change in this PR. 

maybe open one for that change?

kflynn: Sounds like a plan. 🙂

hawkw:
@kflynn 

> One final note: in some places you talk about e.g. "the `HTTPRoute` blah blah" and in other places it's "the HTTPRoute blah blah". We should probably pick one way and go with it. (IIRC the K8s standard is not to set type names in monospace, so that would be my vote, but it's not a terribly strong opinion...)

I tried to keep the use of monospace or not consistent with other existing uses within the same file. I didn't use monospace in the new docs I added, but I also didn't change any existing documentation. It might be worth doing a pass to make this more consistent across all the documentation, but I would prefer to do that in a separate branch.

kflynn:
Yeah, I'm good with limiting scope on this one. Thanks! 🙂

linkerd.io/content/2-edge/features/httproute.md
hawkw:
```suggestion
[HTTPRoute resource]: https://gateway-api.sigs.k8s.io/api-types/httproute/
```

linkerd.io/content/2-edge/features/server-policy.md
hawkw:
```suggestion
  [`Server`]s only (i.e. not [`HTTPRoute`]s)
```

kflynn (approved):
Hit it! 🙂

PR #1660, opened 2023-08-21
stable-2.14.0 upgrade instructions

alpeb:


linkerd.io/content/2-edge/tasks/upgrade.md
```diff
@@ -229,6 +230,14 @@ version.
 
 ## Upgrade notices
 
+### Upgrade notice: stable-2.14.0
+
+For this release, if you're using the multicluster extension, it is mandatory to
+re-link your clusters. `linkerd multicluster check` will error out until you do
```
adleong, line 236: Will re-linking cause any downtime for multicluster services?  Does anything (data plane?) need to be rolled after relinking?  If not, we may want to reassure the user here about those things.

alpeb: This doesn't cause any downtime; I'll add a clarification. No special data plane rollouts required besides what's already described in the top sections.

adleong (approved):

linkerd.io/content/2-edge/tasks/upgrade.md
```diff
@@ -229,6 +230,14 @@ version.
...
+so. Note that this shouldn't cause any downtime in cross-cluster connections.
```
hawkw, line 237: nit, take it or leave it: how about just:
```suggestion
so. This should not cause any downtime in cross-cluster connections.
```

linkerd.io/content/2-edge/tasks/upgrade.md
hawkw, line 236: should this maybe say:
```suggestion
re-link your clusters after upgrading to stable-2.14.0. `linkerd multicluster check` will error out until you do
```

hawkw, line 236: are there consequences of not re-linking besides an error in `linkerd multicluster check`? If so, can we document those reasons it's necessary, as well? Or, if not, can we state that a failed `linkerd multicluster check` is the only consequence of not re-linking?

hawkw (changes requested):

linkerd.io/content/2-edge/tasks/upgrade.md
alpeb: If they don't re-link, the service mirror controllers won't acquire the new images with upgrades like the HA ability we added in this release, like explained above in the "Upgrading the multicluster extension" section. I'll try to reword this a little clearer.

linkerd.io/content/2-edge/tasks/upgrade.md
```diff
@@ -229,6 +230,16 @@ version.
 
 ## Upgrade notices
 
+### Upgrade notice: stable-2.14.0
+
+For this release, if you're using the multicluster extension, you should re-link
+your clusters after upgrading to stable-2.14.0, as explained
+[above](#upgrading-the-multicluster-extension). Not doing immediately so won't
```
hawkw, line 237:
```suggestion
[above](#upgrading-the-multicluster-extension). Not doing so immediately won't
```

linkerd.io/content/2-edge/tasks/upgrade.md
```diff
@@ -229,6 +230,16 @@ version.
...
+cause any downtime in cross-cluster connections, but `linkerd multicluster
+check` will error out.
```
hawkw, line 239: this is nitpicky, but i don't love the phrase "error out". how about
```suggestion
check` will not succeed until the clusters are re-linked.
```

take it or leave it.

hawkw (approved):
looks great!

PR #1662, opened 2023-08-21
document HTTPRoute timeouts

hawkw:
This branch adds new documentation for the HTTPRoute timeout fields
added in Linkerd 2.14. This includes adding the HTTPRoute timeout fields
to the HTTPRoute reference docs, and adding a section on using HTTPRoute
timeouts to the "Configuring Timeouts" tasks doc.

### DIFF
--- a/linkerd.io/content/2-edge/reference/httproute.md
+++ b/linkerd.io/content/2-edge/reference/httproute.md
@@ -58,6 +58,7 @@ HTTPRouteRule defines semantics for matching an HTTP request based on conditions
 | `matches`| A list of [httpRouteMatches](#httproutematch). Each match is independent, i.e. this rule will be matched if **any** one of the matches is satisfied.|
 | `filters`| A list of [httpRouteFilters](#httproutefilter) which will be applied to each request which matches this rule.|
 | `backendRefs`| An array of [HTTPBackendRefs](#httpbackendref) to declare where the traffic should be routed to (only allowed with Service [parentRefs](#parentreference)).|
+| `timeouts` | An optional [httpRouteTimeouts](#httproutetimeouts) object which configures timeouts for requests matching this rule. |
 {{< /table >}}
 
 ### httpRouteMatch
@@ -188,6 +189,31 @@ sent to. Only allowed when a route has Service [parentRefs](#parentReference).
 | `namespace`| Namespace of service for this backend.|
 | `weight`| Proportion of requests sent to this backend.|
 {{< /table >}}
+
+### httpRouteTimeouts
+
+`HTTPRouteTimeouts` defines the timeouts that can be configured for an HTTP
+request.
+
+Linkerd implements HTTPRoute timeouts as described in [GEP-1742]. Timeout
+durations are configured as strings in the format parsed by
+[Go `time.ParseDuration`] (e.g. 1h/1m/1s/1ms), and MUST be greater than 1ms.
+A timeout field with duration 0 disables that timeout.
+
+{{< table >}}
+| field| value |
+|------|-------|
+| `request` | Specifies the duration for processing an HTTP client request after which the proxy will time out if unable to send a response. When this field is unspecified or 0, the proxy will not enforce request timeouts. |
+| `backendRequest` | Specifies a timeout for an individual request from the proxy to a backend service. This covers the time from when the request first starts being sent from the proxy to when the response has been received from the backend. When this field is unspecified or 0, the proxy will not enforce a backend request timeout, but may still enforce the `request` timeout, if one is configured. |
+{{< /table >}}
+
+If retries are enabled, a request received by the proxy may be retried by
+sending it to a different backend. In this case, a new `backendRequest` timeout
+will be started for each retry request, but each retry request will count
+against the overall `request` timeout.
+
+[GEP-1742]: https://gateway-api.sigs.k8s.io/geps/gep-1742/
+[Go `time.ParseDuration`]: https://pkg.go.dev/time#ParseDuration
 
 ## HTTPRoute Examples
 
--- a/linkerd.io/content/2-edge/reference/httproute.md
+++ b/linkerd.io/content/2-edge/reference/httproute.md
@@ -196,7 +196,7 @@ sent to. Only allowed when a route has Service [parentRefs](#parentReference).
 request.
 
 Linkerd implements HTTPRoute timeouts as described in [GEP-1742]. Timeout
-durations are configured as strings in the format parsed by
+durations are specified as strings in the format parsed by
 [Go `time.ParseDuration`] (e.g. 1h/1m/1s/1ms), and MUST be greater than 1ms.
 A timeout field with duration 0 disables that timeout.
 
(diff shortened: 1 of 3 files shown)


linkerd.io/content/2-edge/tasks/configuring-timeouts.md
```diff
@@ -4,16 +4,76 @@ description = "Configure Linkerd to automatically fail requests that take too lo
...
+HTTPRoute timeouts unless a ServiceProfile already exists for the Service.
+
+## Using HTTPRoutes
+
+Linkerd support timeouts as specified in [GEP-1742], for [outbound
+HTTPRoutes](../../features/httproute/#inbound-and-outbound-httproutes)
+with Service parents.
+
+{{< warning >}}
+Support for [GEP-1742](https://gateway-api.sigs.k8s.io/geps/gep-1742/) has not
```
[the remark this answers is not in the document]
kflynn: That's kind of appalling. Ew.

linkerd.io/content/2-edge/tasks/configuring-timeouts.md
```diff
@@ -4,16 +4,76 @@ description = "Configure Linkerd to automatically fail requests that take too lo
...
+will cancel the request and return a [504 Gateway Timeout] response.
+
+Timeouts can be specified either [using HTTPRoutes](#using-httproutes) or [using
+legacy ServiceProfiles](#using-serviceprofiles). Since [HTTPRoute] is a newer
+configuration mechanism intended to replace [ServiceProfile]s, prefer the use of
+HTTPRoute timeouts unless a ServiceProfile already exists for the Service.
+
+## Using HTTPRoutes
+
+Linkerd support timeouts as specified in [GEP-1742], for [outbound
```
kflynn, line 19: s/support/supports/

linkerd.io/content/2-edge/reference/httproute.md
```diff
@@ -189,6 +190,31 @@ sent to. Only allowed when a route has Service [parentRefs](#parentReference).
...
 
+### httpRouteTimeouts
+
+`HTTPRouteTimeouts` defines the timeouts that can be configured for an HTTP
+request.
+
+Linkerd implements HTTPRoute timeouts as described in [GEP-1742]. Timeout
+durations are specified as strings in the format parsed by
+[Go `time.ParseDuration`] (e.g. 1h/1m/1s/1ms), and MUST be greater than 1ms.
+A timeout field with duration 0 disables that timeout.
```
kflynn: GEP-1742 timeouts actually use the syntax specified in GEP-2257, which isn't (quite) the Golang `time.ParseDuration` format. 

I'm not sure how, or if, we want to call this out? In a perfect world, we'd make timeout parsing conform to GEP-2257 now, but I suppose we could also say "yeah, once 1742 goes GA our parsing will change".

linkerd.io/content/2-edge/tasks/configuring-timeouts.md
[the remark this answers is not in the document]
hawkw: yeah, it's not great! although it makes sense, i suppose, since `{{< warning >}}` and friends are Hugo-specific, and markdownlint doesn't know about them...

linkerd.io/content/2-edge/reference/httproute.md
hawkw: This language was copied directly from GEP-1742, which still currently states that the `time.ParseDuration` format is used. Since `time.ParseDuration` is a superset of GEP-2257 (afaik?) it might be best for us to just state that we use that format --- we will actually parse all durations supported by the Go format, but if we document that we use GEP-2257, nobody has to know that we will currently also parse the weird decimal format allowed by Go...

WDYT?

kflynn: I think that'll work, yeah. I do think we should tighten the implementation.

(The language currently in GEP-1742 precedes GEP-2257 -- it's actually what prompted the debate that led to GEP-2257 existing. https://github.com/kubernetes-sigs/gateway-api/pull/2316 updates the 1742 language; it hasn't landed yet because I need to update the API definition stuff first. 🤦‍♂️)

kflynn: (And yes, `time.ParseDuration` is a strict superset of GEP-2257 Durations.)

hawkw: cool, changed it to point to GEP-2257 in https://github.com/linkerd/website/pull/1662/commits/d5131f85302dcc77b9e37f7495f511ab57ce3389. i definitely agree that we should change the implementation to reject the floating-point form eventually.

as a side note, GEP-2257 documents the format substantially more clearly than the Go docs, anyway, so i'm much happier linking to it! :)

linkerd.io/content/2-edge/reference/httproute.md
```diff
@@ -196,8 +196,8 @@ sent to. Only allowed when a route has Service [parentRefs](#parentReference).
 request.
 
 Linkerd implements HTTPRoute timeouts as described in [GEP-1742]. Timeout
-durations are specified as strings in the format parsed by
-[Go `time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) (e.g.
+durations are specified as strings using the [Gateway API duration format]
+specified by [GEP-2257](https://gateway-api.sigs.k8s.io/geps/gep-2257/) (e.g.
 1h/1m/1s/1ms), and MUST be greater than 1ms. A timeout field with duration 0
```
kflynn, line 201: D'oh, just noticed this -- GEP-1742 says >= 1ms rather than > 1ms. Sorry for not catching that before! 🤦‍♂️

kflynn (changes requested):
Sorry! I just noticed one more nit. New syntax language looks great. 🙂

linkerd.io/content/2-edge/reference/httproute.md
hawkw: whoops, will fix!

linkerd.io/content/2-edge/reference/httproute.md
```diff
@@ -189,6 +190,32 @@ sent to. Only allowed when a route has Service [parentRefs](#parentReference).
...
 
+### httpRouteTimeouts
+
+`HTTPRouteTimeouts` defines the timeouts that can be configured for an HTTP
+request.
+
+Linkerd implements HTTPRoute timeouts as described in [GEP-1742]. Timeout
+durations are specified as strings using the [Gateway API duration format]
+specified by [GEP-2257](https://gateway-api.sigs.k8s.io/geps/gep-2257/) (e.g.
+1h/1m/1s/1ms), and MUST be at least 1ms or greater. A timeout field with
```
kflynn, line 202: Nit: "at least" implies "or greater". You don't really need both. 🙂

kflynn (approved):
One wording nit but looks good! Approving whether that gets fixed or not. 🙂

linkerd.io/content/2-edge/reference/httproute.md
hawkw: i felt like this was clearer, since it seemed like there was the potential for someone just quickly skimming the docs to see "MUST be 1ms" and incorrectly interpret it as "all timeouts are 1ms" if they didn't see "or greater". but, that's probably silly...

kflynn (approved):
Hit it!! 🙂

PR #1667, opened 2023-08-22
document Linkerd and Gateway HTTPRoute API groups

hawkw:
As of Linkerd 2.14, we now support HTTPRoutes with both the `policy.linkerd.io` and `gateway.networking.k8s.io` API groups. This branch adds documentation discussing the differences between these two API groups, and notes in the docs stating that both are supported.

### DIFF
--- a/linkerd.io/content/2-edge/features/httproute.md
+++ b/linkerd.io/content/2-edge/features/httproute.md
@@ -9,6 +9,21 @@ aliases = [
 To configure routing behavior and policy for HTTP traffic, Linkerd supports the
 [HTTPRoute resource], defined by the Kubernetes [Gateway API].
 
+{{< note >}}
+Two versions of the HTTPRoute resource may be used with Linkerd:
+
+- The upstream version provided by the Gateway API, with the
+  `gateway.networking.k8s.io` API group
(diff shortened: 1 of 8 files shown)


kflynn (approved):
This looks great. 🙂 Do you think it's worth calling out the need to fully qualify one or the other HTTPRoute when using `kubectl`?

Approving irrespective of the above. 🙂

alpeb:
> This looks great. 🙂 Do you think it's worth calling out the need to fully qualify one or the other HTTPRoute when using `kubectl`?

This is a great point. It appears `httproutes.policy.linkerd.io` now becomes the default, and I remember someone was confused because the `httproutes.gateway.networking.k8s.io` resources weren't showing up.

hawkw:
> > This looks great. 🙂 Do you think it's worth calling out the need to fully qualify one or the other HTTPRoute when using `kubectl`?
> 
> This is a great point. It appears `httproutes.policy.linkerd.io` now becomes the default, and I remember someone was confused because the `httproutes.gateway.networking.k8s.io` resources weren't showing up.

i'll add something about that, good call.

adleong (approved):

PR #1673, opened 2023-08-28
feat: add ngrok ingress controller to ingress guide in docs

russorat:
PM from ngrok here. We have recently released an open source ngrok ingress controller and would love to add it as an option in your guide. 

I was able to run through the getting started guide with linkerd using a local k3d cluster. After completing that guide, i was able to install the ngrok ingress controller into it's own namespace and apply the included ingress.yaml example with no issues. 

I'm not quite sure about the meshing mode needed here. I applied the [example command](https://linkerd.io/2.14/tasks/adding-your-service/#examples) to the ngrok-ingress namespace and didn't have any issues with the emojivoto app, and the linkerd viz dashboard seems to think everything is meshed and working 🤷🏼 

image

### DIFF
--- a/linkerd.io/content/2-edge/tasks/using-ingress.md
+++ b/linkerd.io/content/2-edge/tasks/using-ingress.md
@@ -79,6 +79,7 @@ Common ingress options that Linkerd has been used with include:
 - [Kong](#kong)
 - [Haproxy](#haproxy)
 - [EnRoute](#enroute)
+- [ngrok](#ngrok)
 
 For a quick start guide to using a particular ingress, please visit the section
 for that ingress below. If your ingress is not on that list, never fear—it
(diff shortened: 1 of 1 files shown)


kflynn:
Hey @russorat, thanks for the PR!! The question is, does ngrok's ingress controller send requests using the DNS name of a Service, or does it look up endpoints and go directly to endpoints?

If the former, you _should not_ use Linkerd's ingress mode. If the latter, you _must_ use ingress mode, but you _must_ also inject an `l5d-dst-override` header containing the name of the Service you used to look up the endpoint.

Happy to talk about this more on [Slack](https://slack.linkerd.io/) if that'll help -- thanks again! 🙂

russorat:
great! thanks for the explanation. I verified with the engineering team that our ingress controller uses the DNS name of the service, so it seems as though it can be meshed normally as I have it in the PR.

kflynn (approved):
Many thanks, @russorat! Looks good from here.

PR #1685, opened 2023-09-20
Added enterprise page

travisbeckham:
Added new enterprise page and updated top nav.
With this change, /support-training/ will redirect to /enterprise/


kflynn (approved):
Looks good from here, thanks! 🙂

PR #1687, opened 2023-09-23
Update flagger tutorial

adleong:
Update flagger tutorial to use HTTPRoute instead of TrafficSplit.  Add Argo Rollouts tutorial as well.


kflynn (approved):
Awesome, thanks!! 🙂

PR #1689, opened 2023-09-27
Upgraded to latest Hugo version

travisbeckham:
- Updated .URL to .RelPermalink
- Switched from Blackfriday to Goldmark markdown renderer
- Fixed tweet shortcode
- Fixed highlighting in main nav
- Set Hugo version to 0.119.0


travisbeckham:
Note to reviewers:

The http://linkerd.io/2/... redirect to the latest version is not working in the preview. However, that never works locally for me. Not sure how to test that I didn’t break that functionality.

One of the jobs failed because it can’t find the Hugo config file. That config file was renamed as part of the upgrade. I'm assuming the website-builder container is still using the old Hugo version. If so, how can I get website-builder:v1.3.3 to use Hugo 0.119.0?

wmorgan:
Thank you for tackling this. You've exposed a problem with our build process which still relies on an ancient website-builder image which contains hugo 0.61. It looks like the Dockerfile that builds this image was removed by @olix0r in 38187d15fa1d7a0009ec4adcdce6c3bf2aa9ed69. I'll discuss with him what the most expedient path is going forward.

wmorgan:
The most expedient path forward is to revive the Dockerfile and build a new image of website-builder with an updated version of Hugo.

kflynn (approved):
Many thanks for doing this!! Let's get this landed. 🙂

PR #1691, opened 2023-10-02
Updated runbook link, fixed tweet parameter

travisbeckham:


kflynn (approved):
🤦‍♂️ Thanks! Hit it. 🙂

PR #1699, opened 2023-10-20
remove references to production toolkit

wmorgan:
Let's keep it simple and point them to content we already have on the site.


travisbeckham (approved):
Looks great!

PR #1701, opened 2023-10-26
Add supported-versions doc

kflynn:
Fixes [#11474](https://github.com/linkerd/linkerd2/issues/11474)

### DIFF
--- a/linkerd.io/content/2.14/getting-started/_index.md
+++ b/linkerd.io/content/2.14/getting-started/_index.md
@@ -89,6 +89,10 @@ You should see the CLI version, and also `Server version: unavailable`. This is
 because you haven't installed the control plane on your cluster. Don't
 worry&mdash;we'll fix that soon enough.
 
+Make sure that your Linkerd version and Kubernetes version are compatible by
+checking Linkerd's [supported Kubernetes
+versions](../reference/k8s-versions/).
+
 ## Step 2: Validate your Kubernetes cluster
 
 Kubernetes clusters can be configured in many different ways. Before we can
--- a/linkerd.io/content/2.14/reference/k8s-versions.md
+++ b/linkerd.io/content/2.14/reference/k8s-versions.md
@@ -1,0 +1,22 @@
++++
+title = "Supported Kubernetes Versions"
+description = "Reference documentation for which Linkerd version supports which Kubernetes version"
++++
+
+Linkerd supports all versions of Kubernetes that were supported at the time
+that a given Linkerd version ships. For example, at the time that Linkerd
+stable-2.14.0 shipped, Kubernetes versions 1.26, 1.27, and 1.28 were
+supported, so Linkerd stable-2.14.0 supports all of those Kubernetes versions.
+
+Obviously, Linkerd stable-2.14.0 has no knowledge of what changes will come
+_after_ Kubernetes 1.28. In some cases, later versions of Kubernetes end up
+making changes that cause older versions of Linkerd to not work: we will
+update the chart below as these situations arise.
+
+| Linkerd Version | Minimum Kubernetes Version | Maximum Kubernetes Version |
+|-----------------|----------------------------|----------------------------|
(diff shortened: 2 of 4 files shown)


linkerd.io/content/2.14/getting-started/_index.md
```diff
@@ -85,6 +85,10 @@ You should see the CLI version, and also `Server version: unavailable`. This is
 because you haven't installed the control plane on your cluster. Don't
 worry&mdash;we'll fix that soon enough.
 
+Make sure that your Linkerd version and Kubernetes version are compatible by
```
wmorgan, line 88: I would also add a pointer to this to the upgrading guide.

logicminds:
These are incorrect.  Any version below linkerd-2.13 is not supported by k8s 1.25.  

K8s 1.25 removed many of the deprecated API calls that linkerd uses.

linkerd.io/content/2.14/reference/k8s-versions.md
```diff
@@ -0,0 +1,27 @@
...
+
+Obviously, Linkerd stable-2.14.0 has no knowledge of what changes will come
+_after_ Kubernetes 1.28. In some cases, later versions of Kubernetes end up
+making changes that cause older versions of Linkerd to not work: we will
+update the chart below as these situations arise.
+
+| Linkerd Version | Minimum Kubernetes Version | Maximum Kubernetes Version |
+|-----------------|----------------------------|----------------------------|
+| `stable-2.10`   | `1.21`                     | `1.25`                     |
+| `stable-2.11`   | `1.21`                     | `1.25`                     |
```
logicminds, line 21: 2.10-2.12 has max support of k8s 1.24 only.

kflynn: Thanks!

linkerd.io/content/2.14/getting-started/_index.md
kflynn: 👍

linkerd.io/content/2.14/reference/k8s-versions.md
```diff
@@ -0,0 +1,29 @@
...
+
+Obviously, Linkerd stable-2.14.0 has no knowledge of what changes will come
+_after_ Kubernetes 1.28. In some cases, later versions of Kubernetes end up
+making changes that cause older versions of Linkerd to not work: we will
+update the chart below as these situations arise.
+
+| Linkerd Version | Minimum Kubernetes Version | Maximum Kubernetes Version |
+|-----------------|----------------------------|----------------------------|
+| `stable-2.10`   | `1.16`                     | `1.24`                     |
+| `stable-2.11`   | `1.17`                     | `1.24`                     |
```
alpeb, line 21: I think we should apply this implicitly for core extensions as well (viz, jaeger and multicluster), in which case the maximum version for stable-2.10 and stable-2.11 should be 1.23, given that in 1.24 the automated creation of the token secret for service accounts was removed, which impacted `linkerd mc link` in these two versions.

logicminds: I can confirm multicluster and viz work under 1.24 on an existing cluster.  However, for new clusters the issue above exists so it would be helpful for the user to choose 1.23 then upgrade to 1.24+.   I tried to make 1.24 work on a new cluster but the service account could never find the token I created manually despite following the instructions.

logicminds: Maybe just leave at 1.24 with a special note detailing this issue.

kflynn: I'm actually going to say that their maximum version 1.23. Installing on 1.23 and upgrading to 1.24 is fairly risky, since if you end up needing to reinstall at some point, you'll be in some trouble. 🙂 Let's stick with the conservative route.

alpeb (approved):

PR #1705, opened 2023-11-05
Add troubleshooting section to pod-to-pod-multicluster for required metadata in secret

Majkel1999:
Adding `Troubleshooting` section to the pod-to-pod-multicluster task, as the required metadata for secrets in control plane namespace is not documented.

References:
* https://github.com/linkerd/linkerd2/issues/11542
* https://github.com/linkerd/linkerd2/blob/ce950d17c60eab8c7c2a3e67c130241f96056ac2/controller/api/destination/watcher/cluster_store.go#L63C1-L67

Signed-off-by: @Majkel1999 mkaleta1999@gmail.com

### DIFF
--- a/linkerd.io/content/2.14/tasks/pod-to-pod-multicluster.md
+++ b/linkerd.io/content/2.14/tasks/pod-to-pod-multicluster.md
@@ -305,3 +305,29 @@ default  bb                           authorizationpolicy/bb-authz        0.0rps
 default  default:all-unauthenticated  default/all-unauthenticated         0.0rps  100.00%   0.1rps          1ms          1ms          1ms
 probe    default:all-unauthenticated  default/probe                       0.0rps  100.00%   0.2rps          1ms          1ms          1ms
 ```
+
+## Troubleshooting
+
+### `Failed to get remote cluster`
+
+Multicluster setup requires 2 secrets on the `source` cluster to function correctly. 
+
+1. `cluster-credentials-<remote-cluster-name>` residing in `linkerd-multicluster` namespace (default, it may be different if you have changed it, during installation of `linkerd-multicluster` extension).
+2. `cluster-credentials-<remote-cluster-name>` residing in the LinkerD control-plane namespace -- usually `linkerd`.
+
+The second secret (in `linkerd` namespace) has a specific metadata requirements - `linkerd-destination` required the secret to contain following labels and annotations (assuming we are linking `east` cluster to `west` cluster):
+
+```
+metadata:
+  labels:
+    multicluster.linkerd.io/cluster-name: west
+  annotations:
+    multicluster.linkerd.io/trust-domain: cluster.local
+    multicluster.linkerd.io/cluster-domain: cluster.local
+```
+
+Failure to provide those labels and annotations, causes the `linkerd-destination` to have no way of finding the `kubeconfig` in the secret. It will result in error logs like below:
+
+```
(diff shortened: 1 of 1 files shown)


alpeb:
The `linkerd multicluster link` command should be used to retrieve the Link resource and the other associated resources (including those Secrets) to have multicluster communication properly work. I understand it's not always possible to plug a call to this command in all deployment pipelines, but the manifests to use should still be sourced from this command which provides the ultimate source of truth. So I don't see why we need to single out specific metadata that is already provided, and that is required just like many other pieces of metadata in other linkerd resources.

linkerd.io/content/2.14/tasks/pod-to-pod-multicluster.md
```diff
@@ -305,3 +305,29 @@ default  bb                           authorizationpolicy/bb-authz        0.0rps
...
+## Troubleshooting
+
+### `Failed to get remote cluster`
+
+Multicluster setup requires 2 secrets on the `source` cluster to function correctly. 
+
+1. `cluster-credentials-<remote-cluster-name>` residing in `linkerd-multicluster` namespace (default, it may be different if you have changed it, during installation of `linkerd-multicluster` extension).
+2. `cluster-credentials-<remote-cluster-name>` residing in the LinkerD control-plane namespace -- usually `linkerd`.
+
+The second secret (in `linkerd` namespace) has a specific metadata requirements. Labels and annotations shown below are neccessary for the control plane, to be able to retrieve the remote cluster credentials. Assuming we are linking `east` cluster to `west` cluster, the secret would have to contain below metadata:
```
kflynn, line 318: @Majkel1999, to @alpeb's point, I think it's very important to say here that these Secrets are meant to be managed by the `linkerd multicluster link` command. I'm OK with providing more details, but I feel that the real core thing we need to make sure people understand is that this error generally means that they haven't properly used `linkerd multicluster link`. 🙂

linkerd.io/content/2.14/tasks/pod-to-pod-multicluster.md
```diff
@@ -305,3 +305,29 @@ default  bb                           authorizationpolicy/bb-authz        0.0rps
...
 ```
+
+## Troubleshooting
+
+### `Failed to get remote cluster`
+
+Multicluster setup requires 2 secrets on the `source` cluster to function correctly. 
+
+1. `cluster-credentials-<remote-cluster-name>` residing in `linkerd-multicluster` namespace (default, it may be different if you have changed it, during installation of `linkerd-multicluster` extension).
+2. `cluster-credentials-<remote-cluster-name>` residing in the LinkerD control-plane namespace -- usually `linkerd`.
```
kflynn, line 316: "Linkerd" instead of "LinkerD". 🙂

kflynn (changes requested):
@Majkel1999, thanks for digging into this! I left some comments about how I think this could be improved. 🙂

wmorgan:
I'm going to close this as stale, but I do think the underlying problems here are valid, and should be addressed by some upcoming product work.

ritchellesap:
Hi folks, I'm setting up linkerd multicluster in flat network mode and ran into this issue. I followed this [guide](https://linkerd.io/docs/tasks/pod-to-pod-multicluster/). I'm running on version `edge-26.4.2`.
Can confirm this bug also bit my setup. Adding the metadata labels and annotations was the fix for my linkerd-destination pod. Cheers!

PR #1707, opened 2023-11-08
Update configuration from `linkerd doc`

mateiidavid:
Some of our documentation is extracted from an automatically generated document. This includes (but not limited to) annotation config. Edge documentation has been updated to reflect new protocol detect value

We haven't kept this up to date in a while so this resulted in a larger than expected diff.


alpeb (approved):
Thanks! LGTM!

PR #1708, opened 2023-11-13
Update retry docs.

### LINKED ISSUE #1694: Update "Configuring Retries" page for v2.11 changes
https://linkerd.io/2.14/tasks/configuring-retries/

This page doesn't appear to have been updated for the v2.11 change to allow retries for HTTP routes with bodies and gRPC.

kflynn:
Fixes #1694

### DIFF
--- a/linkerd.io/content/2-edge/tasks/configuring-retries.md
+++ b/linkerd.io/content/2-edge/tasks/configuring-retries.md
@@ -47,9 +47,10 @@ spec:
     isRetryable: true ### ADD THIS LINE ###
 ```
 
-Retries are supported for _all_ idempotent requests, whatever verb they use
-and whether or not they have bodies. In particular, this means that gRPC
-requests can be retried.
+Retries are supported for _all_ idempotent requests, whatever verb they use,
+and whether or not they have a body. In particular, this mean that gRPC
+requests can be retried. However, requests will not be retried if the body
+exceeds 64KiB.
 
 ## Retry Budgets
 
--- a/linkerd.io/content/2-edge/tasks/configuring-retries.md
+++ b/linkerd.io/content/2-edge/tasks/configuring-retries.md
@@ -34,8 +34,8 @@ for a tutorial of how to configure retries.
 
 ## Retries
 
-For routes that are idempotent and don't have bodies, you can edit the service
-profile and add `isRetryable` to the retryable route:
+For routes that are idempotent, you can edit the service profile and add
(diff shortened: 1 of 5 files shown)


linkerd.io/content/2-edge/tasks/configuring-retries.md
```diff
@@ -47,6 +47,10 @@ spec:
     isRetryable: true ### ADD THIS LINE ###
 ```
 
+Retries are supported for _all_ idempotent requests, whatever verb they use
+and whether or not they have bodies. In particular, this means that gRPC
+requests can be retried.
```
hawkw, line 52: we should probably mention the body length limit here, though?

hawkw (commented):
overall, this looks good to me; i think that if we're updating the documentation to reflect that requests with bodies are retryable, we should document the length limit for retryable bodies (64KB; see https://github.com/linkerd/linkerd2-proxy/blob/e62cc28ffe59ba9d1906b04a6873bd95ef23a2f7/linkerd/app/outbound/src/http/retry.rs#L41-L42)

wmorgan (approved):
ty

linkerd.io/content/2-edge/tasks/configuring-retries.md
kflynn: Excellent point. 🙂

linkerd.io/content/2.11/tasks/configuring-retries.md
```diff
@@ -39,6 +39,11 @@ spec:
     isRetryable: true ### ADD THIS LINE ###
 ```
 
+Retries are supported for _all_ idempotent requests, whatever verb they use,
+and whether or not they have a body. In particular, this mean that gRPC
+requests can be retried. However, requests will not be retried if the body
+exceeds 64KiB.
```
hawkw, line 45: do you think we ought to link to the blog post on retry bodies here?

hawkw (approved):
lgtm, one last suggestion

linkerd.io/content/2.11/tasks/configuring-retries.md
kflynn: That's an excellent idea! https://linkerd.io/2021/10/26/how-linkerd-retries-http-requests-with-bodies/ correct?

hawkw: yup!

hawkw (approved):
looks great!

PR #1714, opened 2023-11-21
Adds helm chart to app version matrix

logicminds:
Previously, there was not an easy way to correlate the app version to a helm chart version.  Many people ask this question in public slack channel or must get data themselves.  This adds a script to produce the desired matrix data as a table, json, or yaml output to help folks who are upgrading.  Without this matrix the versions are very confusing.  A python script is provided to generate the necessary data. 

This also adds a new make target to produce the content `build-release-matrix`.  I prepended the target call to the `build-linkerd.io` target so that upon creation the data would be generated.

Screenshot 2023-11-27 at 9 34 34 PM

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ node_modules
 .DS_Store
 tmp
 .idea
-bin
+bin/htmltest
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ serve-api.linkerd.io: build-api.linkerd.io
 		&& python3 -m http.server 9999
 
 .PHONY: build-linkerd.io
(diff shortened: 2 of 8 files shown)


kflynn:
Wow, that looks more painful than I thought. Thanks!! If you can fix the linter errors, I'll see about getting this landed -- appreciate it!

logicminds:
@kflynn please run the checks as I am unable to do so

kflynn (approved):
Let's get this landed. 🙂 Many thanks for sticking with this, @logicminds! I'll likely do some tweaks to the way this fits into the build process, but having this as a base will be a big win.

PR #1733, opened 2024-02-05
Documentation fix: Remove statement about non exiting default timeout for service profiles

mpeuster:
The current documentation states that, if not specified, the default timeout of a route defined via a service profile will be set to 10s.

According to the code and my tests, this default is not in place anymore (see: https://github.com/linkerd/linkerd2/blob/main/controller/api/destination/profile_translator.go#L221). Thus the statement caused some confusion and wrong expectations on our side, so I propose to remove it.


mpeuster:
Anyone interested in this?

alpeb (approved):
Thank you very much @mpeuster for digging into this and testing it :+1:

PR #1737, opened 2024-02-09
update documentation to include canaries when used with SMI interface

miadabrin:
This augments the documentation to provide more context for folks using flagger but still using `ServiceProfile`s (For instance when they need the retry logic).  

This was originally discussed with @kflynn on Slack


miadabrin:
@kflynn it's been a while. are we merging this or should I close it?

wmorgan:
Do we need to copy this to the 2.15 directory as well?

PR #1747, opened 2024-02-28
a few cleanups for edge vs stable releases

wmorgan:
- Removed redundant note about GitHub releases page
- Cluster-specific configuration page mentions Cilium, not Calico, so I updated that. (This page could use some love.)
- Simplified releases page


alpeb (approved):

kflynn (approved):
Taking this one, thanks! 👍

PR #1757, opened 2024-04-03
Updated Automatically Rotating Certs Docs

deusxanima:
Updated automatic cert-manager rotation docs to walk users through process of using cert-manager from start to end. With the new docs we eliminate the need to generate and maintain roots manually with `step` and instead create the entire cert chain with cert-manager. We also introduce using trust-manager to create a `caBundle` and allow linkerd to authenticate the cert-manager certificates without needing to maintain and pass in the ca file manually at install time. 

This change would make the same updates to both identity certificates and webhook certificates.

### DIFF
--- a/linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
+++ b/linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
@@ -6,8 +6,8 @@ aliases = [ "use_external_certs" ]
 
 Linkerd's [automatic mTLS](../../features/automatic-mtls/) feature generates TLS
 certificates for proxies and automatically rotates them without user
-intervention. These certificates are derived from a _trust anchor_, which is
-shared across clusters, and an _issuer certificate_, which is specific to the
+intervention. These certificates are derived from a **trust anchor**, which is
+shared across clusters, and an **issuer certificate**, which is specific to the
 cluster.
 
 While Linkerd automatically rotates the per-proxy TLS certificates, it does not
--- a/linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
+++ b/linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
@@ -6,14 +6,15 @@ aliases = [ "use_external_certs" ]
 
 Linkerd's [automatic mTLS](../../features/automatic-mtls/) feature generates TLS
 certificates for proxies and automatically rotates them without user
-intervention. These certificates are derived from a *trust anchor*, which is
-shared across clusters, and an *issuer certificate*, which is specific to the
+intervention. These certificates are derived from a _trust anchor_, which is
+shared across clusters, and an _issuer certificate_, which is specific to the
 cluster.
 
 While Linkerd automatically rotates the per-proxy TLS certificates, it does not
-rotate the issuer certificate. In this doc, we'll describe how to set up
-automatic rotation of the issuer certificate and its corresponding private key
-using the cert-manager project.
+rotate the issuer certificate. Linkerd's out-of-the-box installations generate
+static self-signed certificates with a validity of one year but require manual
+rotation by the user to prevent expiry. While this setup is convenient for quick
+start testing, it's not advisable nor recommended for production environments.
 
 {{< trylpt >}}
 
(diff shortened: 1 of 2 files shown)


linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -24,17 +25,23 @@ for making TLS credentials from external sources available to Kubernetes
...
 
 Cert-manager is very flexible. You can configure it to pull certificates from
-secrets managemenet solutions such as [Vault](https://www.vaultproject.io).  In
+secrets managemenet solutions such as [Vault](https://www.vaultproject.io). In
 this guide, we'll focus on a self-sufficient setup: we will configure
 cert-manager to act as an on-cluster
 [CA](https://en.wikipedia.org/wiki/Certificate_authority) and have it re-issue
 Linkerd's issuer certificate and private key on a periodic basis, derived from
-the trust anchor.
+the trust anchor. Additionally, we will use trust-manager create a trust bundle
```
kflynn, line 33: "to create"

linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -44,46 +51,111 @@ Linkerd control plane namespace:
...
-kind: Issuer
+kind: Certificate
 metadata:
   name: linkerd-trust-anchor
   namespace: linkerd
+spec:
+  commonName: root.linkerd.cluster.local
+  isCA: true
+  duration: 87600h0m0s
+  renewBefore: 87264h0m0s
```
kflynn, line 127: Huh. I had thought this was "renew, at latest, when this many hours are remaining"?

linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -93,67 +165,131 @@ metadata:
...
-  - cert sign
-  - crl sign
-  - server auth
-  - client auth
+    - cert sign
+    - crl sign
+    - server auth
+    - client auth
+  duration: 28h0m0s
+  renewBefore: 25h0m0s
```
kflynn, line 180: Same question as above re `renewBefore`, but also, `duration` should be 48 hours, correct?

linkerd.io/content/2.15/tasks/automatically-rotating-webhook-tls-credentials.md
```diff
@@ -4,32 +4,33 @@ description = "Use cert-manager to automatically rotate webhook TLS credentials.
...
 ## Install Cert manager
 
-As a first step, [install cert-manager on your
-cluster](https://cert-manager.io/docs/installation/)
-and create  the namespaces that cert-manager will use to store its
+As a first step,
+[install cert-manager on your cluster](https://cert-manager.io/docs/installation/)
+and create the namespaces that cert-manager will use to store its
 webhook-related resources. For simplicity, we suggest using the default
 namespace linkerd uses:
```
kflynn, line 35: Should we refer to the main doc above rather than duplicating the basic setup twice?

linkerd.io/content/2.15/tasks/automatically-rotating-webhook-tls-credentials.md
```diff
@@ -51,66 +52,116 @@ kubectl create namespace linkerd-jaeger
...
-  name: webhook-issuer
-  namespace: linkerd-jaeger
+  name: linkerd-webhook-issuer
+  namespace: linkerd
 spec:
   ca:
-    secretName: webhook-issuer-tls
+    secretName: linkerd-trust-anchor
 EOF
 ```
```
kflynn, line 159: Huh. I've actually been advising people _not_ to use the trust anchor as the root of trust for the webhooks?? They're very different security domains. Have I been contradicting y'all the whole time? :joy:

kflynn (commented):
I have questions, though not about the quality of the writing. 😂

linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
deusxanima: you're correct. updating to clarify.

linkerd.io/content/2.15/tasks/automatically-rotating-webhook-tls-credentials.md
deusxanima: I'm open to it either way. I originally did it this way b/c we have some customers who only set up one and not the other (in both directions), so I wanted both be standalones, but if you think it makes more sense not to duplicate the instructions I'm more than happy to just ref the mTLS doc

linkerd.io/content/2.15/tasks/automatically-rotating-webhook-tls-credentials.md
deusxanima: haha, this is a good question. Overall I think your recommendation is the better one and one that better aligns with proper security practices, even if we're just talking about east-west traffic. My reasoning is really more along the lines of "if someone is already so far in your system that they have access to your private trust anchor keys, you've got bigger problems than two issuers using a single trust anchor", but I do think for security best practices and credibility your position makes more sense. I'll update :)

linkerd.io/content/2.15/tasks/automatically-rotating-control-plane-tls-credentials.md
deusxanima: just double checked their docs: https://cert-manager.io/docs/usage/certificate/#reissuance-triggered-by-expiry-renewal

It looks like the original blurb we had on the site originally (where I copied this blurb from) is accurate: "If `spec.renewBefore` has been set, it will be `spec.renewBefore` amount of time before expiry. cert-manager will set Certificate's `status.RenewalTime` to the time when the renewal will be attempted."

wmorgan:
Revisiting this finally. It appears this is a solution to setting up Linkerd's TLS infra purely through cert-manager, without needing `step`. Given that we recommend cert-manager basically 100% of the time for production deployments, that is compelling, as it allows us to remove an additional tool.

@kflynn Can you please validate that what I said above is true? (And if so, then we will need to decide whether this *replaces* our existing TLS docs, or whether it is an alternative path, or something else.) Thank you. (And thanks @deusxanima!)

kflynn:
I used this as the basis of #1926 -- thanks @deusxanima!!

PR #1768, opened 2024-04-24
Replace kubectl's `--prune-whitelist` with `--prune-allowlist`

alpeb:
The former is no longer supported in latest kubectl versions.


kflynn (approved):
Thanks!

PR #1769, opened 2024-04-26
Fix checks redirect in 2.15 docs

travisbeckham:


linkerd.io/content/2.15/checks/index.html
```diff
@@ -6,7 +6,7 @@
     <script type="text/javascript">
     window.onload = function() {
       var hash = window.location.hash;
-      window.location.href = window.location.origin + "/2.14/tasks/troubleshooting/" + hash;
+      window.location.href = window.location.origin + "/2.15/tasks/troubleshooting/" + hash;
```
wmorgan, line 9: Would it make sense to have this be `/2/tasks/troubleshooting/`?

travisbeckham: Yeah. That should be fine. I should probably update all of the old docs to do the same.

wmorgan (commented):
Thanks! Since this is a new type of error I would advocate for adding some new language to the help doc as well. I will push a commit to the branch.

wmorgan (approved):
Looks great, thank you

PR #1773, opened 2024-05-02
What is a service mesh tweaks

CathPag:
A few stylistic tweaks and correction of the FAQ schema


kflynn:
Forcing DCO because the only committees are Buoyant employees.

kflynn (approved):
We should’ve merged this awhile back. 🤦‍♂️

PR #1774, opened 2024-05-07
clarify release and version guidance

wmorgan:


kflynn (approved):
Hit it! even though you're causing me a couple of merge conflicts. 😂

PR #1777, opened 2024-05-16
Add search index

travisbeckham:
Add search index for 2.15 docs

### DIFF
--- a/.devcontainer/on-create.sh
+++ b/.devcontainer/on-create.sh
@@ -5,7 +5,7 @@ set -euo pipefail
 cd $(mktemp -d)
 
 # hugo
-scurl -O https://github.com/gohugoio/hugo/releases/download/v0.121.2/hugo_extended_0.121.2_linux-amd64.deb
+scurl -O https://github.com/gohugoio/hugo/releases/download/v0.126.1/hugo_extended_0.126.1_linux-amd64.deb
 sudo dpkg -i hugo*.deb
 rm hugo*.deb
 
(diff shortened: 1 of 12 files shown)


wmorgan:
So this will create a JSON file designed for programmatic search engine consumption, correct? Do we need to mark this as noindex or anything like that so that it doesn't appear in e.g. Google search results?

travisbeckham:
> So this will create a JSON file designed for programmatic search engine consumption, correct? Do we need to mark this as noindex or anything like that so that it doesn't appear in e.g. Google search results?

Good point. We could add a robots.txt file, but that might be ignored. Seems like the best way to deal with this is to add a header from the web server. Something like:

```
location ~* \.(json)$ {
 add_header X-Robots-Tag "noindex";
}
```

...but I'm a bit fuzzy on how the linkerd docs site actually get served. Want to just start with a robots file?

travisbeckham:
On a related question, we're currently generating these json files for every section of the site. Do you know what they are for? SEO? https://linkerd.io/2.15/index.json

If so, this search index is overriding it. Let me rework this, so the search index get dumped in the root.

wmorgan:
I am not sure what those files are... maybe some Hugo artifact? At any rate, there is already a robots.txt so the easiest approach is to just add a line there.

travisbeckham:
Those json files are intentionally built, not a Hugo artifact, so I'll avoid clobbering them. I'll assume something reads them.

I've added the search index to robots.txt and moved it to the root.

travisbeckham:
Ok, I think it ready for your review again.

I also fixed some deprecated function calls, and bumped the Hugo version.

wmorgan (approved):

PR #1784, opened 2024-06-04
add Versions section

wmorgan:
Add a versions section at the bottom of Releases and Versions. @kflynn at the risk of conflicting again with your PR, what do you think about something like this to capture the mapping between versions and release artifacts?


kflynn (approved):
Ship it! 🙂

PR #1791, opened 2024-06-21
Feature ambassador

karr9:
styles and code are added to highlight an ambassador


kflynn:
@karr9, muchas gracias! Lo único que prefiero sería que, en vez de aplicar los estilos basado en el nombre del embajador, podemos tener un attribute o algo así? Seguro que en el futuro tendrémos mas embajadores emiriti...

-- 

@karr9, many thanks! The only thing I'd like would be if we could have a flag or some such instead of applying the styling based on the name of the ambassador. I'm sure we'll have more ambassadors emeriti in the future...

karr9:
> ..

ok, agregare un atributo nuevo y en base a ese se hará el highlight

karr9:
@kflynn I have added a new property highlight: true or false, only with this is the change made to the card

kflynn (approved):
¡Gracias @karr9! me parece bien. 🙂

PR #1798, opened 2024-07-02
Switch to page.Path instead of file.Page.Path, because the latter breaks horribly for me, although apparently for no one else???

kflynn:
It doesn't make _any sense_ that .page.File.Path works great in production, but fails on my Mac with _exactly the same version of Hugo_... but it does, and the Hugo docs imply that .page.Path is the smarter way to go. So let's do that.

(And many thanks to @travisbeckham for digging into this with me!!)


travisbeckham (approved):
Looks good! This seems to work well.

PR #1811, opened 2024-08-05
Remove references to "the latest stable version"

wmorgan:
Now that edge releases are the standard way of getting Linkerd, there is no reason for us to tag edge docs with a scary message.

This patch also removes an old task doc about using private Docker repos that mentioned stable releases, which had several quality issues and seemed generally not worth keeping around.


travisbeckham (approved):
Looks good!

kflynn (approved):
I like it. 🙂

PR #1826, opened 2024-08-13
Remove 2.16 code tag link from Releases

travisbeckham:


travisbeckham:
We will be adding the `version-2.16` code tag.

PR #1832, opened 2024-08-27
Copy 2.16 docs to 2-edge

adleong:
Many docs updates were added to 2.16 directly rather than being added to 2-edge and then copied to 2.16.  This means that the 2-edge docs have lagged behind and do not include this updates.  We copy 2.16 to 2-edge in order to pick up all these changes.


kflynn (approved):
🤦‍♂️ Thank you!

PR #1848, opened 2024-09-29
update protocol detection doc

wmorgan:
- Clean up and clarify various bits of language
- Add NATS to the list of protocols that need opaque ports
- Instruct that the resource should always be set on the Service resource

In the interests of simplicity, I kept the guidance around setting the annotation at the Service resource relatively simple.

This content probably needs to be moved into a separate Reference doc. I did not do that in this doc.

### DIFF
--- a/linkerd.io/content/2.16/features/protocol-detection.md
+++ b/linkerd.io/content/2.16/features/protocol-detection.md
@@ -97,6 +97,7 @@ traffic. This is true even if the destination is unmeshed, as it controls the
 behavior of meshed clients.
 
 This annotation *must* be set in two places:
+
 1. On the Service receiving the traffic.
 2. On the workload itself (e.g. on the Deployment's Pod spec receiving the
 traffic), or on enclosing namespace, in which it will apply to all workloads in
--- a/linkerd.io/content/2.16/features/protocol-detection.md
+++ b/linkerd.io/content/2.16/features/protocol-detection.md
@@ -20,9 +20,9 @@ If Linkerd *cannot* determine that a connection is using HTTP or HTTP/2,
 Linkerd will proxy the connection as a plain TCP connection, applying
 [mTLS](../automatic-mtls/) and providing byte-level metrics as usual.
 
-(Note that HTTPS calls to or from meshed pods are treated as TCP, not as HTTP.
-Because the client initiates the TLS connection, Linkerd is not be able to
-decrypt the connection to observe the HTTP transactions.)
+(Note that HTTPS calls pods are treated as TCP, not as HTTP. Because the client
+initiates the TLS connection, Linkerd is not be able to decrypt the connection
+to observe the HTTP traffic.)
 
 ## Configuring protocol detection
 
@@ -34,11 +34,13 @@ This section will help you understand how to fix this.
 {{< /note >}}
 
 In some cases, Linkerd's protocol detection will time out because it doesn't see
-any bytes from the client. This situation is commonly encountered when using
-protocols where the server sends data before the client does (such as SMTP) or
-protocols that proactively establish connections without sending data (such as
-Memcache). In this case, the connection will proceed as a TCP connection after a
-10-second protocol detection delay.
+any bytes from the client within the first 10 seconds of connection
+establishment. This is commonly encountered when using protocols where the
+server sends data before the client does (such as SMTP) or protocols that
+proactively establish connections without sending data (such as Memcache). In
+this case, Linkerd will wait for 10 seconds before establishing a connection to
+the server. After this delay, the connection will be esablished and treated by
+Linkerd as a TCP connection.
 
 To avoid this delay, you will need to provide some configuration for Linkerd.
 There are two basic mechanisms for configuring protocol detection: _opaque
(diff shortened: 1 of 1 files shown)


linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -33,45 +33,53 @@ connections, you are likely running into a protocol detection timeout.
 This section will help you understand how to fix this.
 {{< /note >}}
 
-In some cases, Linkerd's protocol detection will time out because it doesn't see
-any bytes from the client. This situation is commonly encountered when using
-protocols where the server sends data before the client does (such as SMTP) or
-protocols that proactively establish connections without sending data (such as
-Memcache). In this case, the connection will proceed as a TCP connection after a
-10-second protocol detection delay.
+To do protocol detection, Linkerd waits for up to 10 seconds to see bytes sent
```
cratelyn, line 36: i appreciate that this is more specific about the time linkerd waits, instead of the previous "_in some cases_" verbiage.

linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -33,45 +33,53 @@ connections, you are likely running into a protocol detection timeout.
...
+connection establishment to determine the protocol, Linkerd will treat the
+connection as an opaque TCP connection and will proceed as normal, establishing
+the connection to the destination and proxying the data.
 
-To avoid this delay, you will need to provide some configuration for Linkerd.
-There are two basic mechanisms for configuring protocol detection: _opaque
-ports_ and _skip ports_:
+In practice, protocol detection timeouts typically happen when the application
+is using a protocol where the server sends data before the client does (such as
+SMTP) or a protocol that proactively establish connections without sending data
```
cratelyn, line 48:
```suggestion
SMTP) or a protocol that proactively establishes connections without sending data
```

nit

linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -33,45 +33,53 @@ connections, you are likely running into a protocol detection timeout.
...
+(such as Memcache). In this case, everything will work, but Linkerd will
+introduce an unnecessary 10 second delay before connection establishment.
+
+To avoid this delay, you can provide some configuration for Linkerd. There are
+two basic mechanisms for configuring protocol detection: _opaque ports_ and
+_skip ports_:
 
 * Opaque ports instruct Linkerd to skip protocol detection and proxy the
   connection as a TCP stream
 * Skip ports bypass the proxy entirely.
```
cratelyn, line 58:
```suggestion
* Opaque ports instruct Linkerd to skip protocol detection and proxy the
  connection as a TCP stream.
* Skip ports bypass the proxy entirely.
```

punctuation consistency nit, predates these changes

linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -33,45 +33,53 @@ connections, you are likely running into a protocol detection timeout.
...
 
-Opaque ports are generally preferred as they allow Linkerd to provide mTLS,
-TCP-level metrics, policy, etc. Skip ports circumvent Linkerd's ability to
-provide security features.
+Opaque ports are generally preferred as they allow Linkerd to still provide
+mTLS, TCP-level metrics, policy, etc. Skip ports, by contrast, create networking
+rules that avoid the proxy entirely, circumventing Linkerd's ability to provide
+security features.
```
cratelyn, line 63:
```suggestion
Opaque ports are generally preferred as they only skip protocol detection, without interfering with Linkerd's ability to provide
mTLS, TCP-level metrics, policy, etc. Skip ports, by contrast, create networking
rules that avoid the proxy entirely, circumventing Linkerd's ability to provide
security features.
```

a "tioli" (take it or leave it) suggestion. this phrasing tries to be a little more explicit about the distinction between the two mechanisms, and why opaque ports are recommended.

cratelyn (approved):
:taco: lovely work @wmorgan, thank you! i left some minor notes and suggestions, but broadly, i'm very happy to see these docs gain some additional clarity.

linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -33,45 +33,53 @@ connections, you are likely running into a protocol detection timeout.
...
+  connection as a TCP stream.
 * Skip ports bypass the proxy entirely.
 
-Opaque ports are generally preferred as they allow Linkerd to provide mTLS,
-TCP-level metrics, policy, etc. Skip ports circumvent Linkerd's ability to
-provide security features.
+Opaque ports are generally preferred as they only skip protocol detection, without interfering with Linkerd's ability to provide
+mTLS, TCP-level metrics, policy, etc. Skip ports, by contrast, create networking
+rules that avoid the proxy entirely, circumventing Linkerd's ability to provide
+security features.
```
kflynn, line 63: Stupid line-length limits...

```suggestion
Opaque ports are generally preferred as they only skip protocol detection,
without interfering with Linkerd's ability to provide mTLS, TCP-level metrics,
policy, etc. Skip ports, by contrast, create networking rules that avoid the
proxy entirely, circumventing Linkerd's ability to provide security features.
```

linkerd.io/content/2.16/features/protocol-detection.md
```diff
@@ -81,27 +89,32 @@ which configuration you need to apply.
...
+traffic. This is true even if the destination is unmeshed, as it controls the
+behavior of meshed clients.
 
-1. On the workload itself, e.g. on the Deployment's Pod spec receiving the traffic.
-1. On the Service receiving the traffic.
-1. On a namespace (in which it will apply to all workloads in the namespace).
-1. In an [authorization policy](../server-policy/) `Server` object's
-   `proxyProtocol` field, in which case it will apply to all pods targeted by that
-   `Server`.
+This annotation *must* be set in two places:
```
kflynn, line 99: Whoa, this is a big semantic change! Is this really true?

olix0r: Yes, when clients discover policies for a service clusterIP and port, we need to be able to know opaqueness at the service level. If the service is a headless, the service-level annotation will be ignored (because discovery cannot be for the clusterIP), but the general advice to annotate services is correct.

kflynn: Specifically, the old version of this doc says that you can set `config.linkerd.io/opaque-ports` on _either_ the Service or the destination pods/namespace, but this version says that you _must_ set it in both places. That's the semantic change I'm asking about.

It makes a lot of sense to me that there are cases that you'd want to annotate the Service; it's very surprising to me if we're changing to saying that you must annotate both the Service _and_ the destination pods/namespace.

olix0r: It is also required to configure the workloads (via Server or annotation) to control inbound behavior when clients are unmeshed or do not target a clusterIP service.

wmorgan: Basically there are situations where one or the other suffice, and thus the "must" is not strictly necessary. But rather than describing the nuances of those situations I opted to just say, you must set it in both places. If you think that's overkill I'm not committed to the choice, so feel free to suggest other language and we can iterate.

PR #1858, opened 2024-10-23
Update description about proxy wait

mikutas:
While reading https://linkerd.io/2.16/tasks/graceful-shutdown/ 
and https://linkerd.io/2.16/reference/proxy-configuration/ , 
I was a little confused about `config.alpha.linkerd.io/proxy-wait-before-exit-seconds`.

So let me suggest new description about it.

I think it's important when SIGTERM is received.
>When the Linkerd proxy sidecar receives this signal, it will **immediately begin a graceful shutdown where it refuses all new requests** and allows existing requests to complete before shutting down.


wmorgan (changes requested):
@mikutas Thanks for this, but these are generated files. The original string is in the linkerd2 repo. Please make a PR for that repo instead:

https://github.com/linkerd/linkerd2/blob/ae98b359a319939788ddccbf61e54852afcf51c7/cli/cmd/doc.go#L269

PR #1860, opened 2024-10-27
linkerd.io rebuild

travisbeckham:
This PR is a rebuild of linkerd.io.

### Why?

The organization of the front-end assets has devolved to the point where adding new pages and updating existing pages can't be done without adding more tech debt. In short, it's not possible to write good css anymore. Additionally, Hugo has made a lot of improvements over the years, so all of the templates have been rewritten to use the latest features and follow best-practices.

### What's changed?

From a visual perspective, not much has changed. I've kept the look of the website mostly the same, only making small changes to the layout when I felt it was an obvious improvement. However, now it will be much easier to make stylistic changes or implement a full redesign.

There will be process changes when adding content. When adding a new blog post, the frontmatter and file structure will be different from what it is today. Process documentation has been outlined in the [README.md](https://github.com/linkerd/website/blob/travis/next/linkerd.io/README.md). One of the biggest changes is that images are now placed in the same directory as the markdown, and you will need to name your images a certain way, so they will be automatically picked up without needing to provide frontmatter params.

When adding Linkerd documentation, the process is mostly the same, but there's less steps to add a new major version, which is also documented in README.md. Some shortcodes, like fig have also been removed and we'll now use markdown syntax.

### Things to note:

- The Linkerd 1.x docs and 2.9 docs have been dropped.
- ~~The linkerd.io/checks/ redirect has been removed since it's no longer referenced.~~ See PR: https://github.com/linkerd/linkerd2/pull/6102
  - Update: This redirect was referenced in the SMA, so it was re-added.
- The linkerd.io/next-steps/ redirect has been removed since it's no longer referenced. See PR: https://github.com/linkerd/website/pull/1149
- The `helm-chart-version-matrix.md` reference page along with `/bin/generate_release_matrix.py` has been removed. The charts are now hard-coded in the 2.14 docs and below since they will no longer change, and we don't need a matrix for edge.
- The linkerd.io/gsoc/ page has been removed.
- The linkerd.io/choose-your-platform/ page has been removed.
- The `/linkerd.io/release-next-version` script has been removed since the website has dropped 1.x documentation 
- All of the aliases that are not referenced in the `linkerd2` repo have been removed, and all internal links have been updated.


wmorgan (approved):
Let's do it! 🚀

kflynn:
🎉 🙂

PR #1862, opened 2024-10-28
Fixed old blog url

travisbeckham:
Updated an old blog url found in https://linkerd.io/2019/02/22/how-we-designed-retries-in-linkerd-2-2/

Also added some missing aliases where the post date didn’t match the url.


kflynn (approved):
Hit it! Many thanks. 🙂

PR #1869, opened 2024-10-31
Reordered frontmatter params in blog

travisbeckham:
The frontmatter params in blog have been reordered:

1. The date is now the first param
2. The `slug` is now next to the `title` so it's easier to spot the differences between the two


kflynn (approved):
I've checked out the first third or so of the files and I'm just gonna trust that this was mechanical and I needn't check them all individually. 😂

travisbeckham:
> I've checked out the first third or so of the files and I'm just gonna trust that this was mechanical and I needn't check them all individually. 😂

Yep, it was done programmatically, but I also checked each one individually.

PR #1871, opened 2024-11-05
Update latest BEL version

travisbeckham:
Updated latest BEL version on the releases page:
https://deploy-preview-1871--linkerdio.netlify.app/releases/


kflynn (approved):
Hit it! 🙂

PR #1872, opened 2024-11-05
Refactor card

travisbeckham:
The card component has been refactored so that it no longer has breakpoint classes to be more consistent with other components, and to provide more flexibility. With this PR, if you want to change the options of the card at a breakpoint, apply a card mixin. For example:

Current:

```html
<div class="card card-md-reverse">...
```

With this PR:

```scss
@media (min-width: breakpoint-min-width("md")) {
  .card {
    @include make-card--reverse;
  }
}
```

```html
<div class="card">...
```

Reviewers: You can verify the new card behavior on the Enterprise page:
https://deploy-preview-1872--linkerdio.netlify.app/enterprise/

### DIFF
--- a/linkerd.io/assets/scss/app/_adopters.scss
+++ b/linkerd.io/assets/scss/app/_adopters.scss
@@ -11,7 +11,7 @@
     gap: spacer(4);
     grid-template-columns: 1fr;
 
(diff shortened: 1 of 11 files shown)


kflynn (approved):
Sure, I guess? 🙂

PR #1873, opened 2024-11-06
apply protocol detection docs update to edge

adleong:
https://github.com/linkerd/website/pull/1848 updated the protocol detection docs but only applied those changes to the 2.16 docs.  This PR copies those changes to the 2-edge docs as well.


kflynn (approved):
Let's get this landed -- many thanksgh pr review --help :slightly_smiling_face:

PR #1876, opened 2024-11-12
Add minimum Hugo version check

travisbeckham:
Added check that will fail the build if the minimum version of Hugo is not met.


kflynn (approved):
Hit it!

PR #1878, opened 2024-11-18
Fix 2 potencial bugs in docs

travisbeckham:
## Bug 1

Currently, if a section in the Docs sidenav only contains child sections and not any regular pages, the submenu will not be displayed for that section. The submenu should still be displayed in this case.

With this PR, the sidenav has been fixed to show a submenu of a section, if it only contains child sections. 

There are currently not any sections in the Docs sidenav that only contain other sections, but I'm fixing this bug just in case we do in the future.

## Bug 2

Currently, the `section-toc` shortcode is incorrectly listing pages that have the `unlisted` param set.

With this PR, unlisted pages will not be shown when using the `section-toc` shortcode. 

There are currently not any index pages that use the `section-toc` shortcode and also have unlisted pages.

### DIFF
--- a/linkerd.io/layouts/docs/_docs-nav.html
+++ b/linkerd.io/layouts/docs/_docs-nav.html
@@ -54,7 +54,7 @@
         <a href="{{ .RelPermalink }}">
           {{ .LinkTitle }}
         </a>
-        {{ if and .IsSection .RegularPages }}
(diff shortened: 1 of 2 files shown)


kflynn (approved):
Looks good from here! 🙂 Sorry for the delay, too.

PR #1880, opened 2024-11-21
Update distributed tracing docs to include otel

sfleen:
The docs are somewhat outdated here, as they blanket recommend OpenCensus over OpenTelemetry.

Now that OpenCensus is sunset and the proxy supports OpenTelemety, the docs have been updated to reflect that.


kflynn (approved):
Ship it! Even if not perfect it's _much_ better than what's there right now. 😂

wmorgan:
Woot!

PR #1886, opened 2024-11-29
API gateway compatability

GTRekter:
This is a first draft of the compatibility of API gateway and the features supported.


GTRekter:
I've used the linkerd HTTP as reference and created a new page for Gateway API HTTPRoute, and added two examples.

wmorgan:
I took this on as #1909 , PTAL

PR #1898, opened 2024-12-05
Inline author data

travisbeckham:
- Added support for inline author data
- Cleaned up unused author data in `data/authors.yaml`
- Ran `hugo --gc` to clean up resources

---

There are now 2 ways to assign an author to a blog post.

1. If the author has more than one blog post, the author data should be defined in `data/authors.yaml`, then the author id can be set in the frontmatter params. For example:

```yaml
params:
  author: william
```

2. **NEW:** If the author is only going to have a single blog post, the author data can be set inline. For example:

```yaml
params:
  author:
    name: Author name
    avatar: avatar.png
```

For inline author data, the avatar image can either be a page resource or a remote image. To use a remote image, simply provide the full url to the image, and Hugo will download and resize it when the site is published. For example:

```yaml
params:
  author:
    name: John Smith
    avatar: https://example.com/avatars/john-smith.png
```


kflynn (approved):
Wow, awesome! Thanks! 🙂

PR #1903, opened 2025-01-02
We don't actually need this cover image given that we have the video below.

kflynn:


travisbeckham (changes requested):
Instead of deleting the cover image, we should just remove the `showCover: true` frontmatter param. This will allow the cover image to be used when the post is shared on social media.

kflynn:
I think using the thumbnail is fine for sharing, but I've removed the now-pointless `showCover`. Thanks for the catch!

travisbeckham (approved):
Looks great!

PR #1905, opened 2025-01-10
docs: remove references to dead metric `tcp_connection_duration_ms`

nicklasmoeller:
This metric was dropped in https://github.com/linkerd/linkerd2-proxy/pull/930

It's still kept around in the 2.10 docs, since it was only removed from 2.11


kflynn:
The content here looks great, but it looks to me like the metric was in fact removed in stable-2.10.0 – @alpeb, can you doublecheck?

nicklasmoeller:
Happy to change things around - I'm basing the version on the upgrade guide mentioning that it's dropped in 2.11 (under other changes, second bullet): https://github.com/linkerd/website/blob/24b6dba0bd8c1698578df28766858b2a3f789cd3/linkerd.io/content/2.15/tasks/upgrade.md#upgrade-notice-stable-2110

alpeb:
It was removed in linkerd/linkerd2-proxy#930, which was included in the proxy v2.135.0, which was indeed picked in stable-2.10.0 as per https://github.com/linkerd/linkerd2/commit/190f2b13833e742cd038700ac3fe1e62daf51e64

nicklasmoeller:
Alright - I'll drop it from 2.10 docs as well, and keep the upgrade notes across all versions unchanged, even though the version attribution there is incorrect - or do you prefer that we get that changed as well?

kflynn:
I'd vote to change it in the upgrade notes as well, if that's easy?

nicklasmoeller:
I've removed all references now - hope you're OK with that. Moving the note around doesn't really make a lot of sense, without altering the content, and making space for it in another paragraph. Also, it just seems to incorrectly have sneaked into 2.11 notes instead of 2.10 - more reasoning in the commit message.

kflynn (approved):
Looks good from here, thanks! 🙂

nicklasmoeller:
> Looks good from here, thanks! 🙂

No prob - all good! Assume you're handling merging whenever ready 👍

kflynn:
Sorry for the delay, @nicklasmoeller! and many thanks for the fix. 🙂

PR #1908, opened 2025-01-13
Invert order TrafficSplit and Emojivoto App

GTRekter:
create namespace and invert emojivoto deployment and trafficsplit


alpeb (commented):
This looks good, but I think it's a good opportunity to clarify why this ordering is required. I suggest you add a callout in the [Service Profiles Feature doc](https://linkerd.io/2.17/features/service-profiles/) explaining that the workload needs to be restarted if a ServiceProfile is created after the pod started (no need to provide further details). You can reference that paragraph from this tutorial.
Also it'd be great to also mention that subtlety in the Warning callout in the [HTTPRoutes feature doc](https://linkerd.io/2.17/features/service-profiles/) :pray:

wmorgan:
@GTRekter This PR contains a bunch of deletions of 2.9 content for some reason. Can you fix that please?

PR #1910, opened 2025-01-15
Docs: Correcting Prometheus scrape configuration for Multi-Cluster.

fragglehunter:
The scrape configuration on the Linkerd website has a mistake in it that prevents service-mirror metrics from being scraped and showing up in the Viz Dashboard.


wmorgan:
@fragglehunter Can you provide some context around this change, and suggest someone who could review it?

fragglehunter:
@wmorgan 

This update is to fix the service-mirror section in the Prometheus scrape configuration and align it to whats include with the `linkerd-viz` install. 

See example here: [ref.](https://github.com/linkerd/linkerd2/blob/main/viz/charts/linkerd-viz/templates/prometheus.yaml#L81)

If using multicluster, the current configuration prevents the Gateway tab showing up in the viz web UI.

@kflynn would be a good reviewer as we have had some back and forth on the bring your own prometheus topic.

PR #1923, opened 2025-02-12
Update 2-edge K8s versions

kflynn:


kflynn:
I'll do a better audit of places where 2.17 and 2-edge disagree in the future, but a user flagged this one...

PR #1924, opened 2025-02-14
initial draft of the security audit blog post

wmorgan:


linkerd.io/content/blog/2025/0217-security-audit/index.md
```diff
@@ -0,0 +1,68 @@
...
+early this year, was performed by [7ASecurity](https://7asecurity.com/), managed
+by the [Open Source Technology Improvement Fund](https://ostif.org/), and funded
+by the [Cloud Native Computing Foundation](https://cncf.io/). As part of
+Linkerd’s commitment to openness, transparency, and security by design, we’ve
+published the unredacted reports in the [Linkerd GitHub
+page](https://github.com/linkerd/linkerd2/tree/main/audits).
+
+This was the third such audit that Linkerd has undergone, and included both
+pentest and whitebox security audit methods. We were happy to collaborate with
+the 7ASecurity team and OSTIF in the performance of this audit, and particular
```
kflynn, line 27:
```suggestion
the 7ASecurity team and OSTIF in the performance of this audit, and particularly
```

alpeb (approved):

linkerd.io/content/blog/2025/0217-security-audit/index.md
```diff
@@ -0,0 +1,68 @@
...
+  Linkerd 2024 Security Audit
+description: |-
+  We're happy to announce the results of Linkerd's 2024 security audit,
+  courtesy of 7ASecurity, the Open Source Technology Improvement Fund (OSTIF),
+  and the Cloud Native Computing Foundation (CNCF).
+keywords: [linkerd, security]
+params:
+  author: william
+  showCover: true
+images: [cover.jpg]
```
travisbeckham, line 13: This line is not needed, since `cover` is used by default.

travisbeckham (approved):
This looks good with one comment on the frontmatter.

Also, when creating blog posts, it's best if you run Hugo locally so all of the images will be generated and checked in to the repo.

linkerd.io/content/blog/2025/0217-security-audit/index.md
```diff
@@ -0,0 +1,68 @@
...
+no software is perfect, even Linkerd, and every architectural decision
+necessarily involves tradeoffs. The point of a security audit is not to produce
+a report card but to find the weak points and provide opportunities to address
+them before they become user-facing vulnerabilities.
+
+The most severe finding in the audit identified an unused, development-time
+script used to generate protobuf bindings, which (if run) was vulnerable to
+remote code execution. While this script was not part of the modern development
+process, we [removed it from the
+repo](https://github.com/linkerd/linkerd2/pull/13459).
```
olix0r, line 50: Ehhhh... I'll defer to you but here's what this issue was:

https://github.com/linkerd/linkerd2/blob/bc8866efeaf7d1ed8c0092f89df62a1c04ac0fb9/bin/protoc#L27

This line checks if 'unzip' is present and if it isn't, instructs users that they could fetch it from that plaintext HTTP sourcefource.net address. My understanding of their findings was that this use of HTTP (not HTTPS) was insecure. But the script never actually executed this insecure connection, it merely printed out a URL that could _potentially_ be insecure.

travisbeckham:
> @travisbeckham To confirm, I should check in the `resources/_gen/image` directory?

> To confirm, I should check in the `resources/_gen/image` directory?

@wmorgan Yes, this is correct. Here's a bit of background on why it works this way...

The `resources/_gen/images` directory is included in source control because this is where Hugo caches processed images. Each time Hugo is built, it resizes blog images to reasonable sizes, creates thumbnails, then saves them in this directory. If we don't include this directory in source control, Hugo would have to regenerate all of the blog images every time the site is published, resulting is slow build times.

The reason why I implemented this process is because we had blog post cover images that were several megabytes in size that were also being used as thumbnail images.

This process does create some irritation though...

Unfortunately, if we don't run `hugo` locally when creating a new blog post, the images will get created at build time and not be added to source control. This means the next PR where the images are generated locally will include them, which is not ideal. 

To add another irritation to this, Hugo has now changed the naming structure of generated images a couple times. This means that we currently have to be all running version v0.131.0 - v0.141.0, otherwise we'll encounter a renaming mess.

If it is cumbersome having Hugo process the blog images, we can certainly reevaluate this. We'd just need to manually ensure all blog images are created within certain guidlines.

wmorgan:
Current plan is to publish this on Tuesday 2/18

PR #1925, opened 2025-02-17
Improve Hugo version check

travisbeckham:
- Improved Hugo version check to include max version
- Updated readme to include more explanation about generated images


linkerd.io/README.md
```diff
@@ -1,14 +1,79 @@
...
 
-## Website assets
+## General development instructions
+
+### Run the linter and checker
+
+```bash
+docker run \
+  --mount type=bind,source="$(pwd)",target=/website --workdir=/website \
+  ghcr.io/linkerd/dev:v44 sh -c ".devcontainer/on-create.sh && make lint check"
```
kflynn, line 10: It makes me very sad that this isn't a `make` or `just` target. 🙂

kflynn (approved):
Looks good, even though copy-and-pasted commands make me more sad than having `make` targets. 🙂

PR #1926, opened 2025-02-19
Update cert-manager docs

kflynn:
Based on @deusxanima's #1757 and a lot of testing to find the places that cert-manager doesn't work how we thought it works. 🙁 

Probably best to do rich-diff review of the first copy of the docs -- all the others are a copy. @fragglehunter, it'd be great to get your read on this too.

**NOTE WELL** that the links to the Buoyant concept docs will be broken until https://github.com/BuoyantIO/buoyantio.github.io/pull/502 lands.

- **updated automatic cert-manager rotation docs to walk users through process of using cert-manager for entire trust chain**
- **fixed style recommendations**
- **fixing linter style conflicts**
- **WIP updates**
- **Update cert-manager instructions**


linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,885 @@ title: Automatically Rotating Control Plane TLS Credentials
...
-certificates for proxies and automatically rotates them without user
-intervention. These certificates are derived from a *trust anchor*, which is
-shared across clusters, and an *issuer certificate*, which is specific to the
-cluster.
-
-While Linkerd automatically rotates the per-proxy TLS certificates, it does not
-rotate the issuer certificate. In this doc, we'll describe how to set up
-automatic rotation of the issuer certificate and its corresponding private key
-using the cert-manager project.
+Linkerd's [automatic mTLS](../../features/automatic-mtls/), like any mTLS
```
wmorgan, line 6: I'm going to nitpick a couple things. Feel free to TIOLI:
- Like any mTLS implementation -- maybe like any TLS implementation
- "from a trust anchor" -- The "from" here feels too strong since AIUI the only involvement by the trust anchor at this point is that it signed the issuer certificate. Maybe "based on" instead?
- You call these "workload certificates" here, but you use per-proxy certificates" below. I would unify these. Per-proxy is probably more accurate. I feel like I've seen "leaf" certificates elsewhere in the docs.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,885 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+[trust-manager]: https://cert-manager.io/docs/trust/trust-manager/
+
+cert-manager is _extremely_ flexible, and much of its configuration depends on
+the specific policies of the organization running it. Rather than attempt to
+provide a comprehensive guide to cert-manager, this document will focus on a
+very simple setup:
+
+- cert-manager will create a self-signed trust anchor;
+
+- cert-manager will use the trust anchor to create Linkerd's identity issuer
```
wmorgan, line 49: Is there anything we can say about where the key material will be stored for each of these certificates? It seems like that is one of the sticking points in any approach.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,885 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+Linkerd's identity controller.
+
+If you need to force the proxies to rotate their certificates immediately,
+just restart the workloads.
+
+{{< /note >}}
+
+### 9. Rotating the trust anchor
+
+Rotating the trust anchor is a bit different, because (as mentioned before)
```
wmorgan, line 746: I would probably put a note in here saying that trust anchor rotation is optional, and the user should understand the exact threat model that it protects against and evaluate the risk against the level of effort involved.

wmorgan (commented):
Thanks for all the work! I left a few comments. I personally don't think it's necessary to backport this guide to 2.14, etc, but maybe @travisbeckham has another option.

kflynn:
(I don't remember why I had this marked as draft.)

travisbeckham:
> Thanks for all the work! I left a few comments. I personally don't think it's necessary to backport this guide to 2.14, etc, but maybe @travisbeckham has another option.

Yeah, I typically do not add new guides to old versions of the docs. I consider them an archive and only update them when there's an error that needs correcting.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,885 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+certificate_ generated by Linkerd itself from a *trust anchor*, which can be
+shared across clusters, and an *identity issuer certificate*, which is
+specific to the cluster.
+
+{{< note >}}
+
+Certificates are one of the most important parts of a secure system based on
+mTLS, but they are also commonly one of the least well documented. For more
+information about certificates in Linkerd, see the Buoyant <a
+href="https://docs.buoyant.io/buoyant-enterprise-linkerd/latest/reference/tls-concepts/">(m)TLS
```
travisbeckham, line 18: Can we use markdown for this link, instead of html?

kflynn: Inside a `{{< note >}}` or other similar thing, Markdown links don't work. If you know how to fix that I would be _delighted_. 🙂

kflynn:
> Yeah, I typically do not add new guides to old versions of the docs. I consider them an archive and only update them when there's an error that needs correcting.

There actually is. 🙁 Our previous docs don't say set `rotationPolicy`, and without that, cert-manager doesn't change the private key when it "rotates" the cert -- it just updates the validity times but leaves the private key the same. (This was a big surprise – I didn't realize you could get cert-manager to do that all, much less that it would be the default!!) So overall, I'd prefer backporting since it's easy.

travisbeckham:
> > Yeah, I typically do not add new guides to old versions of the docs. I consider them an archive and only update them when there's an error that needs correcting.
> 
> There actually is. 🙁 Our previous docs don't say set `rotationPolicy`, and without that, cert-manager doesn't change the private key when it "rotates" the cert -- it just updates the validity times but leaves the private key the same. (This was a big surprise – I didn't realize you could get cert-manager to do that all, much less that it would be the default!!) So overall, I'd prefer backporting since it's easy.

Gotcha! Sounds good.

kflynn:
> Content and steps look good to me. One minor error with the trailing `\` in code block:

😱 Many thanks!! 🙂

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
kflynn: @travisbeckham and I took a quick look at this and found that the root of my issue was a typo. 🤦‍♂️ Thanks, @travisbeckham!

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
kflynn: I went with `workload` since I only used `per-proxy` in one place.

kflynn:
Addressed feedback and split the automatic-webhook-rotation stuff into a separate branch which I'll PR shortly.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,949 @@ title: Automatically Rotating Control Plane TLS Credentials
...
-automatic rotation of the issuer certificate and its corresponding private key
-using the cert-manager project.
+{{< note >}}
+
+Certificates are one of the most important parts of a secure system based on
+mTLS, but they are also commonly one of the least well documented. For more
+information about certificates in Linkerd, see the Buoyant [(m)TLS concepts
+primer][mtls-concepts-primer].
+
+[mtls-concepts-primer]: https://docs.buoyant.io/buoyant-enterprise-linkerd/latest/reference/tls-concepts/
```
travisbeckham, line 20: I would probably go ahead and link to `https://docs.buoyant.io/primers/tls-concepts/`, so we don't have to update this page in the future. Until the primer section is live on the Buoyant docs, we can create an alias from the reference page to point to this url.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,949 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+more complex: though cert-manager can do the heavy lifting for you, the
+rotation will still involve manual intervention, as explained below.
+
+{{< note >}}
+
+cert-manager is _extremely_ flexible, with many different ways to configure
+it. For more information about cert-manager in general and how to approach its
+configuration, see Buoyant's [cert-manager concepts primer].
+
+[cert-manager concepts primer]: https://docs.buoyant.io/buoyant-enterprise-linkerd/latest/reference/cert-manager-concepts/
```
travisbeckham, line 81: Same comment as above, I would link to `https://docs.buoyant.io/primers/cert-manager-concepts/`.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,949 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+
+```bash
+kubectl get secret -n linkerd linkerd-identity-issuer \
+        -o jsonpath='{ .data.tls\.crt }' \
+    | base64 -d | inspect_cert
+```
+
+Here, we should see something like
+
+```text
```
travisbeckham, line 696: You could add text `{class=disable-copy}` here.

linkerd.io/content/2-edge/tasks/automatically-rotating-control-plane-tls-credentials.md
```diff
@@ -3,200 +3,949 @@ title: Automatically Rotating Control Plane TLS Credentials
...
+base64-encoded, so we can start by dumping it directly:
+
+```bash
+kubectl get configmap -n linkerd linkerd-identity-trust-roots \
+        -o jsonpath='{ .data.ca-bundle\.crt }'
+```
+
+This should be a single PEM CERTIFICATE block:
+
+```bash
```
travisbeckham, line 727: You could add text `{class=disable-copy}` here.

PR #1928, opened 2025-02-27
Fix AuthorizationPolicy docs to indicate that targetRef cannot include a namespace

adleong:
An AuthorizationPolicy cannot target a resource in another namespace, as this would allow one namespace to grant access to another's resources.  This is codified in the CRD as the `targetRef` does not have a `namespace` field.  However, our docs incorrect claim that it does have a `namespace` field.

We correct this error by removing references to the non-existent field and clarifying that only the namespace which can be targeted is the local one.


kflynn (approved):
Thanks @adleong! 🙂

PR #1932, opened 2025-03-13
docs: Add documentation on protocol detection and protocol-aware transport

sfleen:


linkerd.io/content/2-edge/features/protocol-detection.md
```diff
@@ -100,6 +100,10 @@ This annotation *must* be set in two places:
 traffic), or on enclosing namespace, in which it will apply to all workloads in
 the namespace.
 
+You can also set the `appProtocol` field on a Service port to `"linkerd.io/opaque"`
```
wmorgan, line 103: Before I get into a detailed review, I think we need a product call here about whether the appProtocol approach is the new, preferred way for avoiding protocol detection timeouts. If so I would rewrite this section to recommend that approach first, and then add a comment to the earlier approach that prior to 2.18, this was our recommendation but it's no longer the preferred approach because of [reason].

This doc is also now past the point where I would split out the details into a separate reference section, but that could potentially be done as a followup PR by someone else.

linkerd.io/content/2-edge/features/protocol-detection.md
```diff
@@ -112,6 +116,28 @@ instead of a Service annotation. Regardless, we suggest annotating the
 Service object for clarity.
 {{< /note >}}
 
+## Declaring a Service port's protocol
+
+When you're getting started with Linkerd, automatic protocol detection is an
+important way to get as much visibility into your traffic as possible. But as
+you get into more advanced use cases, especially when configuring traffic
+policies, it's more important for protocol handling to be predictable. And if
```
wmorgan, line 124: I think we need to make this explanation better. Justifying protocol detection in terms of visibility seems like it's missing the bigger point. And under which situations is it not predictable? 

I would say something like: automatic protocol detection works in the majority of cases, but it runs the risk that if a connection ever takes more than 10 seconds to send enough data, it might not detect the protocol. This could happen in situations where the cluster is overloaded, the proxy is resource constrained, the app is resource constrained, etc. To eliminate this risk, you can ... etc

linkerd.io/content/blog/2025/0000-protocol-aware-transport/index.md
```diff
@@ -0,0 +1,179 @@
+---
+date: 0000-00-00T00:00:00Z
+title: Protocol-Aware Transport
```
wmorgan, line 3: I like the idea of making this into a blog post if we're excited about this feature and the reasons why we shipped it. But if we're going to do that IMO we'll need to add a fair amount of additional context. I'm happy to work with you on this in e.g. a google doc. If you just want to get the docs shipped asap, you may want to split this into a separate PR.

linkerd.io/content/2-edge/features/protocol-detection.md
sfleen: I would say the preferred way for specifically non-headless service ports, since it lets the user tell us if the protocol is HTTP to continue getting things like request and route-level metrics. There's some cases like direct pod-to-pod communication where there's not a good way to set it without the annotation.

I've moved the section on protocol declaration to be above the opaque to make this preference more clear.

linkerd.io/content/2-edge/features/protocol-detection.md
sfleen: I like that framing better, updated.

linkerd.io/content/blog/2025/0000-protocol-aware-transport/index.md
sfleen: I'll admit I was a bit on the fence about including a blog post at all, since we'd probably want to communicate with BEL partners that want this feature directly. Either way I'll move it into another PR and we can decide this later.

linkerd.io/content/2-edge/features/protocol-detection.md
```diff
@@ -83,6 +83,34 @@ which configuration you need to apply.
...
+resource constrained, the app is resource constrained, etc.
+
+To eliminate this risk, you can set the `appProtocol` field on the ports in a
+Service to determine what protocol to use when communicating with that Service port,
+and skip automatic protocol detection entirely.
+
+| `appProtocol`     | Protocol   | Notes |
+|-------------------|------------|-------|
+| linkerd.io/opaque | opaque     |       |
+| linkerd.io/tcp    | opaque     |       |
```
wmorgan, line 101: Can you please clarify in this table whether `tcp` is exactly the same as `opaque` or whether there is some difference?

linkerd.io/content/2-edge/features/protocol-detection.md
```diff
@@ -100,6 +128,11 @@ This annotation *must* be set in two places:
 traffic), or on enclosing namespace, in which it will apply to all workloads in
 the namespace.
 
+{{< note >}}
+For marking a non-headless Service port as opaque, prefer the `appProtocol`
+setting as outlined above.
+{{< /note >}}
+
```
wmorgan, line 135: There is a section about when you need to change NetworkPolicy at the bottom of the doc. Is it also the case that you need to follow those instructions when you take the appProtocol approach? (To some values, or to any values?) If so could you please document that? (If not please ignore.)

wmorgan (commented):
Looking pretty good, thank you. Two last comments for you.

linkerd.io/content/2-edge/features/protocol-detection.md
sfleen: Yep it's exactly the same, added a note to it.

linkerd.io/content/2-edge/features/protocol-detection.md
sfleen: Added a little note that it doesn't matter how a port is marked as opaque, NetworkPolicy should be set regardless.

wmorgan (approved):
Thank you!

PR #1933, opened 2025-03-13
Mermaid support

travisbeckham:
Added Mermaid support for future use.


kflynn (commented):
This will be lovely, hit it! 🙂

kflynn (approved):
Whoops. That should've been an approval, not just a comment. 🤦‍♂️

PR #1935, opened 2025-03-19
Fix typo in the "Distributed tracing with Linkerd" page

GTRekter:
It's treating the config as a string rather than as an object.


kflynn (approved):
🤦‍♂️ Ouch -- thanks for the catch!! 🙂

cratelyn:
@GTRekter i believe you'll need to amend your commit and add a DCO (developer certificate of origin) described [here](https://github.com/linkerd/website/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

you can run something like:

```sh
git amend --signoff --no-edit && git push --force-with-lease
```

from your branch to push a new, amended commit.

travisbeckham:
I think this needs to be fixed in `2-edge` also.

travisbeckham:
Now that the 2.18 docs are merged, we'll need to fix this in the `2.18` section as well.

cratelyn:
i've opened https://github.com/linkerd/website/pull/1961 to carry this over the finish line.

PR #1941, opened 2025-03-27
Document multicluster links changes for 2.18

alpeb:
The following note added to the `installing-multicluster` doc summarizes well the changes:

> We present here a declarative, GitOps-compatible approach to establishing
multicluster links, available starting with Linkerd `v2.18`. In this method, the controllers are integrated into the multicluster extension, allowing you to supply the Link CR and kubeconfig secrets manifests directly, without necessarily depending on the `linkerd multicluster link` command. This differs from earlier versions of Linkerd (pre-`v2.18`), where (in addition to the Link CR and secrets) controller manifests needed to be provided each time a new link was created, requiring the use of the `linkerd multicluster link` command — a process that was less suited to a GitOps workflow.

This PR updates all the docs(*) referring to multicluster so they follow the new model.

Specific notable changes:

- Given that the new extension-managed controller is just called "controller", we're moving away from the "service mirror controller" appelation.
- We have deprecated the `linkerd mc link` command in favor of the `linkerd mc link-gen` command.
- Likewise, the `linkerd mc unlink` command has been deprecated, so its usage in the docs has been replaced with specific instruction on the resources that need to be deleted.

(*) The "multi-cluster communication with StatefulSets" depends on an external repo, so updating that doc is still pending.


wmorgan (approved):
Thank you for this thorough update

PR #1937, opened 2025-03-22
Update gateway API docs

adleong:

PR #1947, opened 2025-04-03
Add task 'Migrating Gateway API Ownership'

alpeb:


linkerd.io/content/2-edge/tasks/gateway-api-ownership-migration.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Migrating Gateway API Ownership
+description: Instructions for migrating from a Linkerd-managed Gateway API to an externally managed Gateway API, ensuring no loss of dependent CRs during the transition.
```
cratelyn, line 3: it seems like we really ought to include a section somewhere about "the transition", and how things are expected to change. i don't see anything about this in the gateway api referenced below, and we don't describe what the breaking change motivating this _is_ until the final paragraph of this page.

could we..


* (1): update the sentence describing "default behavior" in the gateway api page here https://github.com/linkerd/website/blob/main/linkerd.io/content/2-edge/features/gateway-api.md
* (2): add a note that not only is managing this with an external tool _recommended_ today, it will no longer be the default behavior in default 2.19, in the introductory paragraphs here, before we provide options about how to migrate

alpeb: Sorry, I should have noted that this PR builds on top of #1937 which further clarifies the referred gateway-api doc. In any case, I've added a paragraph at the beginning with your second suggestion :+1:

cratelyn: ah! it's helpful to see those changes as well. thanks for helping me connect these dots :slightly_smiling_face:

cratelyn (approved):

kflynn:
I feel like this is probably the doc where we should talk about how to handle getting from v1alpha2 GRPCRoutes to v1 GRPCRoutes?

kflynn:
I think that the content here is good, but I also think that we need to land #1937 first...

linkerd.io/content/2-edge/tasks/gateway-api-ownership-migration.md
cratelyn: marking this as resolved :white_check_mark: thanks @alpeb!

PR #1954, opened 2025-04-15
Remove image processing

travisbeckham:
@kflynn For your consideration. Happy to discuss the pros and cons further.

Hugo image processing has been removed because it has been a struggle to maintain the generated images in source control for a few reasons:

- Hugo has changed the naming format of the generated iamge a couple times, which forces everyone that works on the website to use a narrow range of versions.
- If someone is not using a supported Hugo version, the version check can not prevent image generation even though it can fail the build, which results in all of the generated images for the whole website being added to their PR using incorrect file names.
- Using image processing requires the developer to run Hugo locally so the generated images are associated with the correct PR, and this is easy to forget.

Removing image processing will result in larger image sizes on the blog list page. If this becomes a problem, we should manually scale the cover image down, or created thumbnail images.


kflynn (approved):
Looks good from here! 🙂 I think it's the right call, too; it seems like maintaining objects in the source repo eventually always causes trouble.

PR #1956, opened 2025-04-16
fix minor issues in egress docs

adleong:
* quote annotation value so that it isn't treated as a boolean
* use correct client pod name
* use v1 HTTPRoute resource version


kflynn (approved):
Thank you!! 🙂

PR #1958, opened 2025-04-18
Update protocol detection doc

adleong:
The protocol detection doc was out of date and included incorrect information.  I've updated the doc to reflect current best practices with Linkerd around `appProtocol` and opaque ports.


kflynn (approved):
So much better -- thanks! 🙂

PR #1971, opened 2025-05-01
Document how to re-enable authority metric label

adleong:


linkerd.io/content/2-edge/reference/proxy-metrics.md
```diff
@@ -64,6 +64,32 @@ Each of these metrics has the following labels:
...
+```
+
+This can also be specified on the command line if installing with the CLI:
+
+```shell
+linkerd install \
+  --set 'proxy.additionalEnv[0].name=LINKERD2_PROXY_INBOUND_METRICS_AUTHORITY_LABELS' \
+  --set 'proxy.additionalEnv[0].value=unsafe'
+```
+
```
olix0r, line 92:
```suggestion
```

I'd omit this -- the standard practice would be to use `linkerd install -f values.yaml`. This is obviously a tip if you really want to avoid the config but I don't think it's worth putting in the public docs as a golden path.

olix0r (approved):
not a blocker but i'd omit the verbose CLI usage.

PR #1974, opened 2025-05-07
Truncate summary when used as page description

travisbeckham:
Currently, when the `.Summary` is use for the meta description because the `.Description` value is empty, the summary content is not truncated. Because of this, the length of the summary can be longer than the recommended length of a page description.

With this PR, when the summary is used as the page description, it is truncated to 160 characters.


kflynn (approved):
Looks good! 🙂

PR #1975, opened 2025-05-09
The Story Behind the Great Sidecar Debate

kflynn:


travisbeckham:
I think the cover would be better saved as a jpg instead of a png (479 KB => 74 KB). 

cover

travisbeckham (approved):
Looks great!

PR #1977, opened 2025-05-13
Include full content in search index

travisbeckham:
Improve search result by including the full content of doc in the search results.


kflynn (approved):
Definitely -- thanks! 🙂

PR #1986, opened 2025-07-02
Fix broken blog post links

travisbeckham:
I noticed several broken links in blog posts that bypassed the `htmltest` check because they were prefixed with the linkerd.io domain. Also, because these links contain a domain, they are opening in a new window, which is also not desirable.

For example, see the "promised in our February Linkerd 2.15 announcement" link in [this blog post](https://linkerd.io/2024/08/13/announcing-linkerd-2.16/).

With this PR, the linkerd.io domain has been removed in blog post links and a redirect has been added for `/community/anchor/`.


kflynn (approved):
Ship it! Thanks much. 🙂

PR #1989, opened 2025-07-09
Multicluster Upgrade instructions (only in 2.18)

GTRekter:
I’ve added an extra block on the **Installing Multi-Cluster Components** page that outlines the process for upgrading from versions prior to 2.18.


travisbeckham:
@GTRekter @alpeb I'm wondering if we should just use the same guidance that's in the BEL docs as a standalone Task page?
https://docs.buoyant.io/buoyant-enterprise-linkerd/latest/upgrade/#multicluster-upgrade=multicluster-upgrade-cli&upgrading-the-multicluster-extension

GTRekter:
Just noticed that that page has already a section for that

> This will deploy a new controller for each link (named controller-<link-name>), while leaving the old controllers (named linkerd-service-mirror-<link-name>) in place. The new controllers won’t do anything yet because the old controllers will hold the Lease object for each Link.
> To fully migrate to the new model, delete the old controller deployments and their associated resources, calling for each link:
> linkerd multicluster unlink --cluster-name <link-name> --only-controller |
> 	kubectl delete -f -
> After a few seconds the new controllers will reclaim the Lease object and start managing those links.

travisbeckham:
@GTRekter @alpeb I have an update...

After talking with William about this, we decided it's best to create a new task doc in `edge-2` and `2.18` named `Upgrading Multi-cluster Components`. The content of the page should be roughly the same as the [Upgrading the multicluster extension](https://docs.buoyant.io/buoyant-enterprise-linkerd/latest/upgrade/#upgrading-the-multicluster-extension) section in the BEL docs.

After this doc is published, we will add a link to it from the [edge release notes](https://github.com/linkerd/linkerd2/releases/tag/edge-25.3.3).

@GTRekter Would you like that tackle that change?

PR #1990, opened 2025-07-09
Highlight the importance of the Bundle and the need to restart both CP and DP

GTRekter:
Added two warnings aimed at preventing misunderstandings about cert-manager automatically rotating certificates and clarifying the need for a manual step.


kflynn (changes requested):
Thanks! and sorry for the delay. I've suggested a couple of tweaks, let me know what you think! 🙂

PR #2009, opened 2025-08-04
Remove unnecessary cert-manager RBAC setup from docs

kahirokunn:
cert-manager already has cross-namespace secret creation permissions by default, as shown here:
https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/rbac.yaml#L110-L112

This PR removes the manual RBAC configuration steps from the documentation since they are no longer needed.


kflynn (approved):
Thanks @kahirokunn! and sorry it took so long to verify this. 🤦‍♂️

PR #2012, opened 2025-08-05
Add aspect-ratio to img class

travisbeckham:


kflynn (approved):
Excellent, ship it! 🙂 (Also, TIL... 😂)

PR #2013, opened 2025-08-07
Remove .metadata.namespace from Bundle

kahirokunn:
Bundle is a cluster-level resource, hence .metadata.namespace was removed.

```
$ kubectl api-resources
NAME                                SHORTNAMES   APIVERSION                                NAMESPACED   KIND
bundles                                          trust.cert-manager.io/v1alpha1            false        Bundle
```


kflynn (approved):
Looks good, thank you! 🙂

PR #2015, opened 2025-08-08
Add Linkerd Ambassador

travisbeckham:
- Added Blake Romano as Linkerd Ambassador 
- Optimized Ambassador images


kflynn (approved):
Looks great, thanks! 🙂 and wow, I didn't realize that the dimensions of these headshot images were so wildly varying...

travisbeckham:
> Looks great, thanks! 🙂 and wow, I didn't realize that the dimensions of these headshot images were so wildly varying...

Yeah, when we had Hugo image processing turned on, I converted all of them to a standard size and format. With this PR, I decided not to resize the images, in case we need a larger size in the future, but just converted the pngs to jpgs to get rid of some extra kBs.

PR #2030, opened 2025-08-25
Add reference to the linkerd=debug directive in the Proxy Log Level page

GTRekter:
The Proxy Log Level page only references the `linkerd2_proxy` directive, but it is the `linkerd` directive that provides a lot of information about the overall behavior of the proxy.


travisbeckham:
Completed via #2048

PR #2036, opened 2025-09-05
Doc updates

kflynn:
There are two doc fixes here:

1. Fix the retry documentation to correctly show which annotations can be used on which resources.

2. Take a first pass at calling out SMI as deprecated _now_, rather than in the future.

There's more to do with docs that use TrafficSplits in the examples; that'll be a separate PR later.

### DIFF
--- a/linkerd.io/content/2.18/tasks/automatic-failover.md
+++ b/linkerd.io/content/2.18/tasks/automatic-failover.md
@@ -7,9 +7,9 @@ description: Use the Linkerd Failover extension to failover between clusters.
 
 The Linkerd Failover extension is deprecated and will be removed in a future
 release. If you are currently using this functionality, we recommend migrating
-to use [federated Services](../tasks/federated-services.md) instead, which is
-a much simpler approach to achieving high availability across clusters that
-does not require the Linkerd Failover extension.
+to use [federated services](../tasks/federated-services) instead, which is a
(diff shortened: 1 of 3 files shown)


linkerd.io/content/2.18/tasks/automatic-failover.md
```diff
@@ -3,6 +3,16 @@ title: Automatic Multicluster Failover
 description: Use the Linkerd Failover extension to failover between clusters.
 ---
 
+{{< warning >}}
+
+The Linkerd Failover extension is deprecated and will be removed in a future
+release. If you are currently using this functionality, we recommend migrating
+to use [federated Services](../tasks/federated-services.md) instead, which is
```
travisbeckham, line 10: we can remove the extension from this url: `../tasks/federated-services/`

travisbeckham, line 10: Should "Services" be lowercase? We call it "federated services" in other docs.

travisbeckham (approved):
Just a couple small nits, but this looks good to merge.

linkerd.io/content/2.18/tasks/automatic-failover.md
kflynn: Lowercased for consistency even though I'm partial to the capital S here. 😂

PR #2041, opened 2025-09-15
Add edge-version and chart-version shortcodes

travisbeckham:
This PR introduces 2 new shortcodes: `edge-version` and `chart-version`.

### `edge-version`

`edge-version` will return the edge version for a corresponding Linkerd version based on the current section.

If the current section is not `2.xx`, then the latest edge release will be returned. For example, `/2-edge/` or `/faq/`.

If the current section is `2.15` or higher, then the corresponding edge release for that Linkerd version will be returned. For example, if a doc in the `2.17` section calls `edge-version`, it will return "edge-24.11.8". In the `2.18` section, it will return "edge-25.4.4". 

If you want an edge release for a certain Linkerd version, regardless of the current section, you can pass the Linkerd version as a param. For example, `{{< edge-version "2.18" >}}` will always return "edge-25.4.4".

### `chart-version`

`chart-version` will return the Helm chart version for a corresponding Linkerd version based on the current section using the same logic as `edge-version`.

### Notes

The only visible change with this PR is in the "Recent versions" on the [Releases page](https://deploy-preview-2041--linkerdio.netlify.app/releases/). It now uses a new data file that the new shortcodes also use. The announcement text uses the blog post title, so there's small differences in text from prod.

Also be aware that the `chart-version` shortcode will only work in prod if the “L5D2_EDGE_VERSION” string gets replaced **before** Hugo builds the site, [which I’m pretty sure this is the case](https://github.com/linkerd/website/blob/baa92cc5fb1e06303fdd4e8085e4a38ee4060d02/Makefile#L25). The “L5D2_EDGE_VERSION” string does not get replaced in Netlify previews.

### DIFF
--- a/linkerd.io/layouts/shortcodes/chart-version.html
+++ b/linkerd.io/layouts/shortcodes/chart-version.html
@@ -1,7 +1,19 @@
 {{- /*
 
 `chart-version` will return the Helm chart version for a corresponding Linkerd
-version based on the current section using the same logic as `edge-version`.
+version based on the current section.
+
+If the current section is not "2.xx", then the Helm chart version for the latest
+edge release will be returned.
+
+If the current section is 2.15 or higher, then the corresponding Helm chart
+version for that Linkerd version will be returned. For example, if a doc in the
+2.17 section calls `chart-version`, it will return "2024.11.8". In the 2.18
+section, it will return "2025.4.4".
+
+If you want the Helm chart for a certain Linkerd version, regardless of the
+current section, you can pass the Linkerd version as a param. For example,
+`{{< chart-version "2.18" >}}` will always return "2025.4.4".
 
 */ -}}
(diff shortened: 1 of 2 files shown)


kflynn:
Checking the Makefile, `get-versions` is a dependency of `build-linkerd.io`, so we should be good to go here. 🙂

kflynn:
In the [release page preview](https://deploy-preview-2041--linkerdio.netlify.app/releases/), this doesn't seem quite right?

> The latest edge release is [L5D2_EDGE_VERSION](https://github.com/linkerd/linkerd2/releases/tag/L5D2_EDGE_VERSION). and the full list of edge release artifacts can be found on [the Linkerd GitHub releases page](https://github.com/linkerd/linkerd2/releases).

kflynn:
In that case, ship it! 🙂

kflynn (approved):

PR #2042, opened 2025-09-19
Fix Upgrade Linkerd section in gitops doc

travisbeckham:


kflynn (approved):
After far more checking Helm versions than I wanted to do, this looks good, ship it! 🙂

PR #2101, opened 2026-03-02
fix(docs): remove metadata.namespace from ClusterIssuer

mikutas:
similar to #2013 

ClusterIssuer is also cluster-level resource


alpeb (approved):
Good find! 👍

PR #2102, opened 2026-03-04
Update libsass to dart sass

travisbeckham:
The embedded LibSass transpiler in Hugo was deprecated in v0.153.0 and will be removed in a future release. LibSass was deprecated by the Sass team back in 2020.

This PR upgrades the transpiler to Dart Sass, modernizes the scss files, and updates the build workflow.

To use Dart Sass with Hugo for local development, install Dart Sass somewhere in your PATH, Hugo will find it. For example, to install with Homebrew, run: `brew install sass/sass/sass`.

To use Dart Sass with Hugo when deploying to branch previews or production, the build workflow has been updated to install Dart Sass before the Hugo site build begins.

**Preview**
https://deploy-preview-2102--linkerdio.netlify.app/


kflynn (approved):
I think we go for it as long as the Netlify previews look good. I can't think of anything else we could reasonably do to test.

PR #2112, opened 2026-04-23
Improve CLI reference pages

travisbeckham:
This PR fixes 2 things:

1. On the [cli reference pages](https://linkerd.io/2.19/reference/cli/viz/#subcommands), the headers for Flags and Examples and a few other headers are the wrong level (for instance, Flags is an h2 but the command name is an h3).
2. The visibility of the subcommands is not great. For instance, if you're looking for the `tap` subcommand it can be hard to find.

With this PR:

- Fix header level consistency in cli reference pages
- Add table of subcommand on a command detail page

**Preview**
https://deploy-preview-2112--linkerdio.netlify.app/2.19/reference/cli/viz/#subcommands

### DIFF
--- a/linkerd.io/content/2-edge/reference/cli/diagnostics.md
+++ b/linkerd.io/content/2-edge/reference/cli/diagnostics.md
@@ -10,6 +10,8 @@ title: diagnostics
 
 ## Subcommands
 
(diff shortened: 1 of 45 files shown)


kflynn (approved):
Thanks for the additional review context! Looks good from here. 🙂

PR #2115, opened 2026-05-02
Add /reference redirect to latest version (#1849)

### LINKED ISSUE #1849: `linkerd.io/reference` could redirect to latest reference docs
as of writing, the latest reference documentation lives at https://linkerd.io/2.16/reference/. other versions' documentation can be accessed at `:{version}/reference/`.

it would be nice if https://linkerd.io/reference/ redirected clients to the latest reference, much like how https://linkerd.io/docs redirects to

singhvishalkr:
Issue #1849 reports that `https://linkerd.io/reference` 404s. The versioned reference docs live at `linkerd.io/{version}/reference/`, but a reader needs to already know the version number to find them. The same problem was solved years ago for `/docs` via `linkerd.io/content/docs/_index.md`, which redirects through the Hugo `redirect` layout to `/2/getting-started/` and then onward to `/{latestMajorVersion}/getting-started/`.

This PR adds the equivalent file for `/reference/`. The new `linkerd.io/content/reference/_index.md` reuses the same `_default/redirect` layout with `params.redirect: /2/reference/`. The existing `redirect.html` template (`linkerd.io/layouts/_default/redirect.html`) rewrites `/2/{path}` into `/{latestMajorVersion}/{path}` at render time, so this entry will keep pointing at the current latest reference docs every time a new minor ships, without any further code change.

No existing pages live under `/reference/`, so there is no collision: a `gh api .../contents/linkerd.io/content/reference` lookup currently 404s on `main`.

Fixes #1849

### DIFF
--- a/linkerd.io/content/reference/_index.md
+++ b/linkerd.io/content/reference/_index.md
@@ -1,0 +1,9 @@
+---
+title: Reference
+
+# Redirect
+type: _default
+layout: redirect
+params:
+  redirect: /2/reference/
+---


travisbeckham:
Thank you so much for taking the time to address this. This highlighted a much bigger improvement that needed to be made that would enable permanent urls for all docs in the latest version, not just the reference section.

This improvement has been made in [PR #2116](https://github.com/linkerd/website/pull/2116) , and we now have a permanent url for the [latest reference page](https://linkerd.io/docs/reference/).

PR #2117, opened 2026-05-14
Improve template logic to find subcommands

travisbeckham:
On the current site, to find the subcommands for a certain command we loop through all of the commands looking for a certain prefix.

With this PR, the logic has been optimized to find these subcommands using the `like` operator.

There should not be any visible changes to the site:
https://linkerd.io/docs/reference/cli/multicluster/#subcommands
https://deploy-preview-2117--linkerdio.netlify.app/docs/reference/cli/multicluster/#subcommands


kflynn (approved):
🚢 🙂

PR #2122, opened 2026-06-01
fix: use sass-embedded in local setup

immanuwell:
Problem
Fresh local setup still points people at plain `sass`, but that can break the Hugo build with `TOCSS-DART ... got unexpected EOF when executing "sass"`. Pretty easy to hit.

Fix
Point the README and Makefile hint to `sass-embedded`, which makes the same build pass.

Repro
1. `tmp=$(mktemp -d) && cd "$tmp"`
2. `npm init -y >/dev/null 2>&1 && npm install --save-dev sass >/dev/null 2>&1`
3. `PATH="$tmp/node_modules/.bin:$PATH" hugo -s /path/to/website/linkerd.io --minify --cleanDestinationDir`
4. see the `TOCSS-DART` error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ ifndef HAS_HUGO
 	@printf "Install hugo first. For OSX: brew install hugo\n"; exit 1
 endif
 ifndef HAS_SASS
-	@printf "Install Dart Sass first: npm install sass\n"; exit 1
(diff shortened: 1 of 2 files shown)


travisbeckham (approved):
This looks good to me. Thanks!

PR #2123, opened 2026-06-03
Update edge docs for native sidecars enabled by default in Linkerd 2.20

alpeb:
As of Linkerd 2.20 the proxy is injected by default as a native sidecar container (an init container with `restartPolicy: Always`) instead of a regular container. Update the edge docs accordingly:

- features/native-sidecars.md: state the new default and replace the "Enabling" section with instructions for disabling the mode, via the `config.linkerd.io/proxy-enable-native-sidecar: false` annotation at the namespace or workload level, or globally via the Helm value `proxy.nativeSidecar: false`
- features/proxy-injection.md, reference/architecture.md: mention the proxy is now injected as a native sidecar by default
- features/cni.md, tasks/grafana.md: scope the initContainer networking limitation (and its workarounds) to the case where native sidecars have been disabled, since init containers now have network access by default
- tasks/graceful-shutdown.md: meshed Jobs and Cronjobs now complete without extra configuration; scope the hanging-pod issue and the manual /shutdown workaround to the non-native-sidecar case. Drop the `--wait-before-seconds` option and the "Slow Updating Clients" section, which are no longer relevant with native sidecars
- tasks/adding-your-service.md: fix the mesh verification command to look for the proxy under `spec.initContainers` instead of `spec.containers`

### DIFF
--- a/linkerd.io/content/2-edge/tasks/graceful-shutdown.md
+++ b/linkerd.io/content/2-edge/tasks/graceful-shutdown.md
@@ -100,7 +100,9 @@ which is no longer the default. In that case the proxy runs as a regular
 container that runs continuously until it receives a TERM signal, and since
 Kubernetes does not give the proxy a means to know when the Job has completed,
 meshed Job and Cronjob pods will continue to run even once the main container
-has completed. You can address this by manually shutting down the proxy.
+has completed. You can address this by manually shutting down the proxy, after
+the application container completes. This triggers a graceful shutdown allowing
+meshed Job and Cronjob pods to complete.
 
 ### Manual shutdown
 
--- a/linkerd.io/content/2-edge/tasks/graceful-shutdown.md
+++ b/linkerd.io/content/2-edge/tasks/graceful-shutdown.md
@@ -66,12 +66,6 @@ signal.
 Linkerd offers a few options to configure pods and containers to gracefully
 shutdown.
 
-- `--wait-before-seconds`: can be used as an install value (either through the
-  CLI or through Helm), or alternatively, through a
-  [configuration annotation](../reference/proxy-configuration/). This will add a
-  `preStop` hook to the proxy container to delay its handling of the TERM
-  signal. This will only work when the conditions described above are satisfied
-  (i.e container runtime sends the TERM signal)
 - `config.linkerd.io/shutdown-grace-period`: is an annotation that can be used
   on workloads to configure the graceful shutdown time for the _proxy_. If the
   period elapses before the proxy has had a chance to gracefully shut itself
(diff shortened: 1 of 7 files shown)


linkerd.io/content/2-edge/tasks/graceful-shutdown.md
```diff
@@ -92,68 +86,26 @@ shutdown.
...
+shuts down when the main containers in the pod terminate, so meshed Jobs and
+Cronjobs complete without any extra configuration.
+
+This is only an issue if you have [disabled native
+sidecars](../features/native-sidecars/#disabling-native-sidecars-in-linkerd),
+which is no longer the default. In that case the proxy runs as a regular
+container that runs continuously until it receives a TERM signal, and since
+Kubernetes does not give the proxy a means to know when the Job has completed,
+meshed Job and Cronjob pods will continue to run even once the main container
+has completed. You can address this by manually shutting down the proxy.
```
raykroeker: > You can address this by manually shutting down the proxy.

It's not clear what the sequence here should be.  Maybe:

You can address this by manually shutting down the proxy, after the application container completes. This triggers a graceful shutdown allowing meshed Job and Cronjob pods to complete.

linkerd.io/content/2-edge/tasks/grafana.md
```diff
@@ -58,11 +58,14 @@ instructions on how to easily import the same charts published on
...
-outbound port 443 via `config.linkerd.io/skip-outbound-ports: "443"` annotation
-or run the container with the proxy's UID. See
+configuration and dashboards. If you use the CNI plugin and have
+[disabled native sidecars](../features/native-sidecars/#disabling-native-sidecars-in-linkerd)
+(which are enabled by default as of Linkerd 2.20), when you add grafana's pod
+into the mesh its initContainer will run before the proxy is started and the
+traffic cannot flow. In that case you should either avoid meshing grafana's
+pod, skip outbound port 443 via the
+`config.linkerd.io/skip-outbound-ports: "443"` annotation, or run the container
+with the proxy's UID. See
```
raykroeker, line 68: Running multiple processes under the same UID feels like a general anti-pattern.  I understand we recommend it elsewhere, should we continue to promote it?

raykroeker (approved):
I left a comment related to clarity around manual shutdown and a question.

linkerd.io/content/2-edge/tasks/graceful-shutdown.md
alpeb: 👍

linkerd.io/content/2-edge/tasks/grafana.md
alpeb: One advantage of this approach is that it keeps that specific init container out of the mesh, whereas the annotation alternative would apply to all containers in the pod. That said, I can see your point; it may be worth discussing further. If you feel strongly about it, please feel free to open a ticket so we can track it

PR #2127, opened 2026-06-12
fix: avoid copying cert-manager labels/annotations to linkerd-previous-anchor

mikutas:
The previous sed/egrep pipeline copied cert-manager labels/annotations to linkerd-previous-anchor, causing cainjector to log "unable to fetch certificate that owns the secret".

Replaced it with a `jq` filter that reconstructs only the necessary fields, so no cert-manager labels or annotations are carried over.


alpeb (approved):
Thanks @mikutas , this looks good to me! 👍

PR #2134, opened 2026-06-24
Truncate opengraph and twitter card description to 160 chars

travisbeckham:
For the meta description on the site, we truncate to 160 characters. However, we are currently using Hugo’s built templates for opengraph and twitter meta data, and they do not truncate the description. When the frontmatter description is not set, the templates fall back to using the content summary which is truncated at 70 words (too long).

With this PR, the opengraph and twitter card templates have been pulled into the site, so the description will be truncated to 160 characters. A `metaTitle` frontmatter param has also been added, so a page can optionally set the meta title to something different than the page title.

The Hugo version has been bumped to v0.158.0 to support `Language.Locale`.

**Preview**
https://deploy-preview-2134--linkerdio.netlify.app/2026/06/24/federating-clusters-for-zero-downtime-kubernetes/


CathPag (approved):
LGTM

alpeb (approved):

PR #2139, opened 2026-06-28
fix: remove dead links from blog posts

immanuwell:
Fixes 3 dead links in older blog posts.

Why
These source pages are live, but the links inside them send readers to 404s right now. Pretty easy to hit in practice, just click through.

Repro
`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2020/07/23/under-the-hood-of-linkerds-state-of-the-art-rust-proxy-linkerd2-proxy/`
`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2016/06/17/small-memory-jvm-techniques-for-microservice-sidecars/`

`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2021/07/28/announcing-cncf-graduation/`
`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2016/02/18/linkerd-twitter-style-operability-for-microservices/`

`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2022/12/28/service-mesh-2022-recap-ebpf-gateway-api/`
`curl -s -o /dev/null -w '%{http_code}\n' https://linkerd.io/2022/12/01/what-really-happens-at-startup-linkerd-init-containers-the-cni-and-more/`

What changed
Removed 2 dead links where there is no matching in-repo target, and fixed 1 slug to the live 2022 post. nice and boring.

Checks
`make build-linkerd.io`
`make lint`

### DIFF
--- a/linkerd.io/content/blog/2020/0723-under-the-hood-of-linkerd-s-state-of-the-art-rust-proxy-linkerd2-proxy/index.md
+++ b/linkerd.io/content/blog/2020/0723-under-the-hood-of-linkerd-s-state-of-the-art-rust-proxy-linkerd2-proxy/index.md
@@ -72,10 +72,9 @@ ran on the Java Virtual Machine, it had a pretty significant resource footprint.
 (The JVM is good at "scaling up", but not good at "scaling _down_", as William
 wrote in his InfoQ article on
 [the decision to reimplement Linkerd](https://www.infoq.com/articles/linkerd-v2-production-adoption/).)
-Even though the Linkerd community
-[got pretty good](/2016/06/17/small-memory-jvm-techniques-for-microservice-sidecars/)
-at tuning the JVM's memory use to minimize footprint, it was still too much to
-ask in a per-pod service mesh deployment model. So we knew we'd need a language
+Even though the Linkerd community got pretty good at tuning the JVM's memory use
+to minimize footprint, it was still too much to ask in a per-pod service mesh
+deployment model. So we knew we'd need a language
 that compiled to native binaries, like Rust, Go, and C++.
 
 Now, latency. Another lesson we learned from Linkerd 1.x that informed our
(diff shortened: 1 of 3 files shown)


travisbeckham:
Thanks for catching these!

What I don't understand is why `htmltest` did not catch these in the `validate.yml` workflow? Any ideas?

immanuwell:
@travisbeckham I dug into this
validate.yml is running htmltest, but htmltest is configured with IgnoreURLs: `^/2/*`
Because IgnoreURLs uses regexes, that pattern matches any internal path beginning with /2, including dated blog permalinks like /2016/... and /2022/..., so these broken links were skipped by CI
I verified this locally by reproducing it with the repo's exact .htmltest.yml: htmltest passes on a broken /2016/... link, and fails as soon as that ignore rule is removed or tightened
So the root cause is an over-broad ignore regex in .htmltest.yml, not the workflow itself

travisbeckham:
> @travisbeckham I dug into this validate.yml is running htmltest, but htmltest is configured with IgnoreURLs: `^/2/*` Because IgnoreURLs uses regexes, that pattern matches any internal path beginning with /2, including dated blog permalinks like /2016/... and /2022/..., so these broken links were skipped by CI I verified this locally by reproducing it with the repo's exact .htmltest.yml: htmltest passes on a broken /2016/... link, and fails as soon as that ignore rule is removed or tightened So the root cause is an over-broad ignore regex in .htmltest.yml, not the workflow itself

Oh, great catch! Thanks for doing the investigation. Now that we're no longer using `/2/` internally, it's probably safe to remove it from `IgnoreURLs` in a follow up PR.

travisbeckham (approved):
Thanks for fixing these broken links, and for figuring out why `htmltest` didn't catch them.

PR #2142, opened 2026-06-30
Fix lang on command output examples

travisbeckham:
Currently, there are a lot of code blocks in the troubleshooting page that show examples of command output. These 
code blocks have the Bash language set, so the syntax highlighting doesn't make sense in a lot of cases. These code blocks also do not need a Copy button.

For example:
Screenshot 2026-06-30 at 2 27 09 PM

This PR changes language of code blocks on command output examples from Bash to text, and disables the Copy button.

**Current**
https://linkerd.io/docs/tasks/troubleshooting/#kubectl-version

**Preview**
https://deploy-preview-2142--linkerdio.netlify.app/docs/tasks/troubleshooting/#kubectl-version

### DIFF
--- a/linkerd.io/content/2-edge/tasks/replacing_expired_certificates.md
+++ b/linkerd.io/content/2-edge/tasks/replacing_expired_certificates.md
@@ -16,7 +16,7 @@ replace the expired certificates with valid certificates.
 It might be the case that your issuer certificate is expired. If this it true
 running `linkerd check --proxy` will produce output similar to:
 
-```bash
+```text {class=disable-copy}
 linkerd-identity
 ----------------
 √ certificate config is valid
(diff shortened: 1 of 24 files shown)


alpeb (approved):
LGTM! 👍 
Would be wonderful at some point to be able to have the copy button only select the relevant part of the snippet, there were we have both the command and the output 😉 😉

travisbeckham:
> LGTM! 👍 Would be wonderful at some point to be able to have the copy button only select the relevant part of the snippet, there were we have both the command and the output 😉 😉

Yep. I'm currently working on a new shortcode that does this! My current prototype looks like this:
Screenshot 2026-06-30 at 4 50 12 PM

Once I get it in good shape, I'll get it added to this site.

PR #2148, opened 2026-07-24
Remove query parser characters from search term

travisbeckham:
Lunr's query parser treats characters like : + - * ~ ^ as query syntax, so this PR removes them to keep user input literal and avoid throwing an error if a search term contains one of them.

**Lunr docs**
https://lunrjs.com/guides/searching.html#fields

> The search term is prefixed with the name of the field, followed by a colon (:). The field must be one of the fields defined when building the index. **Unrecognised fields will lead to an error.**


alpeb (approved):
I wasn't aware of that syntax. It could give us finer query control, but I guess there's a trade-off with ease-of-use and predictability that argues against enabling these advanced features in this context.

PR #2151, opened 2026-08-13
docs(proxy-metrics): add section about metrics expiry

anthoturc:
this commit adds documentation around metric expiry in the linkerd proxy.

some users may have seen that metrics are discarded after 10 minutes. this gotcha wasn't documented previously and has lead to confusion.

the docs in this commit are applied to the edge version of linkerd because the behavior has been around since at least 2.10. 

this addresses an ask in linkerd/linkerd2#6222

### DIFF
--- a/linkerd.io/content/2-edge/reference/proxy-metrics.md
+++ b/linkerd.io/content/2-edge/reference/proxy-metrics.md
@@ -370,3 +370,38 @@ following labels:
   https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label
 [ttfb]: https://en.wikipedia.org/wiki/Time_to_first_byte
 [HTTPRoute]: httproute/
+
+## Metric Expiry
+
+The Linkerd proxy emits some metrics that have high-cardinality labels. For example,
+clusters with a large number of meshed nodes may have many IP addresses for the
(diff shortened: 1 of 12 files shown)


cratelyn (commented):
i like this change, it's a great idea to document this property of these legacy metrics.

> the docs in this commit are applied to the edge version of linkerd because the behavior has been around since at least 2.10.

if this affects proxies going back to at least 2.10, what do you think of applying this same change to the page in [2.19](https://github.com/linkerd/website/blob/main/linkerd.io/content/2.19/reference/proxy-metrics.md), [2.18](https://github.com/linkerd/website/blob/main/linkerd.io/content/2.18/reference/proxy-metrics.md), etc.?

anthoturc:
> if this affects proxies going back to at least 2.10, what do you think of applying this same change to the page in [2.19](https://github.com/linkerd/website/blob/main/linkerd.io/content/2.19/reference/proxy-metrics.md), [2.18](https://github.com/linkerd/website/blob/main/linkerd.io/content/2.18/reference/proxy-metrics.md), etc.?

I'm not opposed! I can update those as well.

cratelyn (approved):

cratelyn:
nice! thank you very much :pray:

travisbeckham (approved):
These changes look great!

PR #2153, opened 2026-09-04
fix(install): make binary install atomic

### LINKED ISSUE #90: run.linkerd.io/install can corrupt installs
While trying to explain to someone that run.linkerd.io/install safely installs the linkerd binary without potentially corrupting an existing one, i realized there's a subtle problem: Our `mv` may not be atomic.

https://github.com/linkerd/website/blob/6d29e3a58a82838f3114db3aae338aa51769c3d8/run.linkerd.io/public/install#L13

On most modern OSes /tmp and $HOME exist on different devices, so that

https://github.com/linkerd/website/blob/6d29e3a58a82838f3114db3aae338aa51769c3d8/run.linkerd.io/public/install#L34-L37

is not guaranteed to be atomic. So if a user hits `^C` during an install, they can be left with a `linkerd` binary that is partial.

If we change the `mktemp` to use a temp directory in the `$HOME/.linkerd2`, we can ensure that the `mv` is atomic; however, this would mean that the installer would have to be responsible for garbage-collecting these tmp directories.

I suppose it could also be okay to close this as wontfix if we provided safe alternatives (i.e. via homebrew, apt-get, etc).

amarkdotdev:
## Summary
- Create install temp directories under `$INSTALLROOT` (default `~/.linkerd2`) so `mv` to `$INSTALLROOT/bin` stays on the same filesystem and remains atomic.
- Garbage-collect leftover `linkerd2.*` temp directories from interrupted installs before starting a new download.

Fixes #90

## Test plan
- [ ] Run `curl -fsSL https://run.linkerd.io/install | sh` (or invoke the script locally) and confirm the CLI installs successfully
- [ ] Interrupt an install mid-download and re-run; confirm stale temp dirs are removed and install completes
- [ ] Verify `linkerd version` works after install


amarkdotdev:
Just nudging for maintainer feedback on this PR.

PR #2155, opened 2026-09-13
docs: remove duplicate 'a' in AuthorizationPolicy authentication note

mrchatam:
Fixes #2154

Remove duplicate article in the AuthorizationPolicy note:

- before: "require a a variety of forms of authentication"
- after: "require a variety of forms of authentication"

Docs-only, one-line change.

PR #2157, opened 2026-09-21
fix(configuring-per-route-policy): Remove duplicate 'a'

travisbeckham:


kflynn (approved):
Thanks! and thanks to @mrchatam for the original find in #2155. Ship it!

